PR #8, opened 2017-04-21
bundle.bbclass: allow variables in RAUC_SLOT_ varflag [file]

ejoerns:
This will expand the [file] varflag given for a RAUC_SLOT_ and thus
allows using variables in them, such as

    RAUC_SLOT_rootfs = "my-${MACHINE}-rootfs.ext4"

### DIFF
--- a/classes/bundle.bbclass
+++ b/classes/bundle.bbclass
@@ -24,7 +24,7 @@
 #
 #   RAUC_SLOT_dtb ?= linux-yocto
 #   RAUC_SLOT_dtb[type] ?= "file"
-#   RAUC_SLOT_dtb[file] ?= "am335x-bone.dtb"
+#   RAUC_SLOT_dtb[file] ?= "${MACHINE}.dtb"
 #
 #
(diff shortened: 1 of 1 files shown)


srn-skov-dk:
Can we have a comment added in the bbclass file that give a short intro to this usage.
As it is now you almost needs to be python expert to realise this flexibility exists
/Sam

ejoerns:
To be honest, I would expect this flexibility to exist, but it didn't. What we could do is giving a hint this works in one of the examples by using `${MACHINE}` in it. Nevertheless, I just saw the commit message is wrong (should be `RAUC_SLOT_rootfs[file]`) thus I will have to update this PR, anyway.

ejoerns:
I've now fixed the commit message and added using `${MACHINE}` to the example in the classes description:

``` diff
-#   RAUC_SLOT_dtb[file] ?= "am335x-bone.dtb"
+#   RAUC_SLOT_dtb[file] ?= "${MACHINE}.dtb"
```

PR #10, opened 2017-05-24
add support for RAUC hawkBit client

ejoerns:
This adds recipes for building the (just published) [rauc-hawkbit](https://github.com/rauc/rauc-hawkbit) client.

This is an initial version of the recipes, neither heavily tested nor equipped with any service startup infrastructure. Feel free to test it, provide feedback and improve it.

In order to build it, you also need to enable meta-openembedded/meta-python and meta-openembedded/meta-oe in you bblayers.conf.

### DIFF
--- a/README
+++ b/README
@@ -20,6 +20,12 @@ This layer depends on:
   layers: meta-oe
   branch: master
 
+For rauc-hawkbit client:
+
+  URI: git://git.yoctoproject.org/meta-oe
+  layers: meta-python
+  branch: master
+
 
 Patches
 =======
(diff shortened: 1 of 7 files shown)


ejoerns:
Updated to some more recent package versions and added required packages

skny-prevas:
Thanks for adding the exact same fixes and version bumps as i did last weekend :-)
I have tested this and it works on my setup 👍 
How is your thoughts on rebooting targets after an update? Are you adding a reboot command to the post-handle in the bundle?

ejoerns:
Sounds good! If we both did the same independently from each other, it cannot be totally wrong ;)

In our Demo where we use the Code atm, we do the rebooting by the applilcation that executes the update, because we show a reboot countdown to possibly interrupt reboot for testing or demonstration purpose.
If you simply want to reboot after a successful installation, you can use a post-install handler for this.

RAUC also supports (in the near future) scenarios where activating an installation and rebooting is done remote-controlled, e.g. at a certain scheduled point in time, or when all devices out in the field have downloaded and installed their update. In general, it mainly depends on your needs what fits best.

ejoerns:
updated to build rauc-hawkbit 0.1.0 and added description to README

PR #12, opened 2017-06-06
rauc: allow override CA certificate file entry in SRC_URI

hundeboll:
Append files to the rauc recipe from other layers might need to use a
CA certificate external to the build directory, in which case it should
be possible to use an absolute path in the SRC_URI.

In case of certificate files not named ca.cert.pem as expected by the
rauc recipe, append files can use the `downloadfilename`, e.g.

RAUC_CA_FILE = "file://path/to/our/ca.crt;downloadfilename=ca.cert.pem"

### DIFF
--- a/recipes-core/rauc/rauc-target.inc
+++ b/recipes-core/rauc/rauc-target.inc
@@ -1,6 +1,7 @@
+RAUC_KEYRING_FILE ??= "file://ca.cert.pem"
 SRC_URI_append = " \
   file://system.conf \
-  file://ca.cert.pem \
+  ${RAUC_KEYRING_FILE} \
   file://rauc-mark-good.service \
   "
 


recipes-core/rauc/rauc-target.inc
```diff
@@ -1,6 +1,7 @@
+RAUC_CA_FILE ??= "file://ca.cert.pem"
```
ejoerns: Please use the name `RAUC_KEYRING_URI` for the variable (in the recipe and description message) as this better reflects the purpose and ability of this variable. I would also add a blank line here.

ejoerns:
Could you perform the following change (and an appropriate adaption of commit message) on your patch?

``` diff
diff --git a/recipes-core/rauc/rauc-target.inc b/recipes-core/rauc/rauc-target.inc
index 6873512..1d3ce1c 100644
--- a/recipes-core/rauc/rauc-target.inc
+++ b/recipes-core/rauc/rauc-target.inc
@@ -1,7 +1,8 @@
-RAUC_CA_FILE ??= "file://ca.cert.pem"
+RAUC_KEYRING_FILE ??= "file://ca.cert.pem"
+
 SRC_URI_append = " \
   file://system.conf \
-  ${RAUC_CA_FILE} \
+  ${RAUC_KEYRING_FILE} \
   file://rauc-mark-good.service \
   "
 ```

Additionally, please also add your Signed-off-by line to the patch as it is described in the `rauc` base project https://github.com/rauc/rauc/blob/master/docs/contributing.rst#developers-certificate-of-origin.

mnhu:
Updated.

ejoerns:
I just fixed the commit message (still contained `RAUC_CA_FILE`) and merged it manually, then.

PR #16, opened 2017-08-15
rauc-native: use DEPLOY_DIR_TOOLS

jremmet:
switching machines result in

ERROR: rauc-native-0+gitAUTOINC+3867722b8d-r9 do_deploy: The recipe
rauc-native is trying to install files into a shared area when those
files already exist. Those files and their manifest location are:
   build/deploy/tools/rauc-0+gitAUTOINC+3867722b8d
 Matched in b'manifest-MACHINE1-rauc-native.deploy'
 build/deploy/tools/rauc
 Matched in b'manifest-MACHINE2-rauc-native.deploy'
Please verify which recipe should provide the above files.

Using DEPLOY_DIR_TOOLS in do_deploy fixes this.

Signed-off-by: Jan Remmet <j.remmet@phytec.de>

PR #18, opened 2017-09-19
rauc-native: append datetime to tooldir filename

hundeboll:
When rauc-native is rebuilt due to changes others than ${PV}, the
do_deploy step fails because of ${DEPLOY_DIR_TOOLS}/rauc-${PV} already
being present.

Fix this by appending ${DATETIME} to the install filename, and avoid
basehash/taskhash mismatches by excluding ${DATETIME} from the
basehash/taskhash.

Signed-off-by: Martin Hundebøll <mnhu@prevas.dk>

### DIFF
--- a/recipes-core/rauc/rauc-native_git.bb
+++ b/recipes-core/rauc/rauc-native_git.bb
@@ -2,12 +2,15 @@ require rauc-git.inc
 
 inherit native deploy
 
+RAUC_TOOL_BASENAME = "rauc-${PV}-${DATETIME}"
+RAUC_TOOL_BASENAME[vardepsexclude] = "DATETIME"
+
 do_deploy[sstate-outputdirs] = "${DEPLOY_DIR_TOOLS}"
 
 do_deploy() {
     install -d ${DEPLOY_DIR_TOOLS}
-    install -m 0755 ${B}/rauc ${DEPLOY_DIR_TOOLS}/rauc-${PV}
-    ln -sf rauc-${PV} ${DEPLOY_DIR_TOOLS}/rauc
+    install -m 0755 ${B}/rauc ${DEPLOY_DIR_TOOLS}/${RAUC_TOOL_BASENAME}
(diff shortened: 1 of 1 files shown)


ejoerns:
This is a different solution for the issue targeted by the just-merged PR #16. Did you take a look at this, before? I hope it also matches your needs?

hundeboll:
Our CI has experienced the same issue with #16 merged (building with meta-rauc commit 7f5c3e024747ec32f5b61c8aea8fd5fc4da152ed):
```
ERROR: rauc-native-0.1.1+gitAUTOINC+4b6453e867-r9 do_deploy: The recipe rauc-native is trying to install files into a shared area when those files already exist. Those files and their manifest location are:
  /home/buildbot/slave/builders/machine-beagleboneblack/build/yocto/tmp-glibc/deploy/tools/rauc-0.1.1+gitAUTOINC+4b6453e867
    (not matched to any task)
Please verify which recipe should provide the above files.
```

I don't understand why changing from
```
install -m 0755 ${B}/rauc ${DEPLOYDIR}/rauc-${PV}
```
to
```
install -m 0755 ${B}/rauc ${DEPLOY_DIR_TOOLS}/rauc-${PV}
```
should fix it, since both approaches end up installing into `do_deploy[sstate-outputdirs]` with identical filenames.

ejoerns:
Yes, but I assumed that bypassing the additional copy from sstate would prevent this check to be triggered. I tested it locally and it looked ok, and I also took a look at what the meta-oe dfu-util-native recipe does, which is quite the same:
https://github.com/openembedded/meta-openembedded/blob/master/meta-oe/recipes-support/dfu-util/dfu-util-native_0.9.bb

If that really does not solve the issue, I fear you approach (which is of course quite the same what image recipes do) is required to fix this.

mnhu:
When trying to reproduce on my local machine, the fix in #16 seems to work, so I guess I just have to clean `${DEPLOY_DIR_TOOLS}` on the CI slaves...

You can close this, thanks.

ejoerns:
Sounds good. If we should across this issue again, we could reopen this PR.

PR #19, opened 2017-09-19
classes/bundle: depend on do_deploy task for kernel and bootloader

### LINKED ISSUE #14: Missing "do_image_complete" for kernel
This is a "repost", since I did not see a response in the mailing list:

I’m attempting to add a kernel to my rauc bundle, based off of phytec’s “linux-ti” kernel: https://git.phytec.de/meta-phytec/tree/recipes-kernel/linux/linux-ti_4.4.52-phy2.bb  However, when I add the kernel, I’m getting an error `bundle.bb depends upon non-existent task do_image_complete`.  It appears that the only class which contains `do_image_complete` is the image class, but as far as I understood, the kernel shouldn’t be a system image.  I am able to create a file with just the file system, but adding a kernel or dts file does not work.

Looking inside the bundle.bbclass file, it appears that it is using the `do_image_complete` task for fetching dependencies.  Am I supposed to create this task for my kernel somehow?  What is it looking for as a return value?

Below is my bundle.bb file:

```
inherit bundle

RAUC_KEY_FILE = "${THISDIR}/development-1.key.pem"
RAUC_CERT_FILE = "${THISDIR}/development-1.cert.pem"
RAUC_BUNDLE_COMPATIBLE= "277264"

RAUC_BUNDLE_SLOTS="rootfs kernel"

RAUC_SLOT_rootfs="phytec-qtdemo"
RAUC_SLOT_rootfs[fstype]="ubifs"

RAUC_SLOT_kernel="linux-ti"
RAUC_SLOT_kernel[type]="kernel"
```

ejoerns:
Depending on `do_image_complete` task is valid for images but
inappropriate for kernel and bootloader which do not have such a task.

This changes the dependency auto-generation in order to honor the image
`type` varflag provided and depend on the `do_deploy` task for kernel
and bootloader.

This fixes #14

### DIFF
--- a/classes/bundle.bbclass
+++ b/classes/bundle.bbclass
@@ -58,8 +58,16 @@ RAUC_BUNDLE_BUILD       ??= "${DATETIME}"
 RAUC_BUNDLE_BUILD[vardepsexclude] = "DATETIME"
 
 # Create dependency list from images
-do_fetch[depends] = "${@' '.join([d.getVar(image) + ":do_image_complete" for image in \
-    ['RAUC_SLOT_' + slot for slot in d.getVar('RAUC_BUNDLE_SLOTS').split()]])}"
+python __anonymous() {
+    for slot in (d.getVar('RAUC_BUNDLE_SLOTS') or "").split():
+        slotflags = d.getVarFlags('RAUC_SLOT_%s' % slot)
+        imgtype = slotflags.get('type') if slotflags else None
+        image = d.getVar('RAUC_SLOT_%s' % slot)
+        if imgtype == 'kernel' or imgtype == 'boot':
+            d.appendVarFlag('do_fetch', 'depends', ' ' + image + ':do_deploy')
+        else:
+            d.appendVarFlag('do_fetch', 'depends', ' ' + image + ':do_image_complete')
+}
 
 S = "${WORKDIR}"
 
--- a/classes/bundle.bbclass
+++ b/classes/bundle.bbclass
@@ -36,7 +36,7 @@ LICENSE = "MIT"
 
 PACKAGE_ARCH = "${MACHINE_ARCH}"
 
-RAUC_IMAGE_FSTYPE ??= "${@(d.getVar('IMAGE_FSTYPES', True) or "").split()[0]}"
+RAUC_IMAGE_FSTYPE ??= "${@(d.getVar('IMAGE_FSTYPES') or "").split()[0]}"
 
 do_fetch[cleandirs] = "${S}"
 do_patch[noexec] = "1"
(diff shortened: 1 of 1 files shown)


classes/bundle.bbclass
```diff
@@ -58,8 +58,16 @@ RAUC_BUNDLE_BUILD       ??= "${DATETIME}"
 RAUC_BUNDLE_BUILD[vardepsexclude] = "DATETIME"
 
 # Create dependency list from images
-do_fetch[depends] = "${@' '.join([d.getVar(image, True) + ":do_image_complete" for image in \
-    ['RAUC_SLOT_' + slot for slot in d.getVar('RAUC_BUNDLE_SLOTS', True).split()]])}"
+python __anonymous() {
+    for slot in d.getVar('RAUC_BUNDLE_SLOTS', True).split():
```
mnhu: No need to pass `True` to `d.getVar()`, as this is the default since November 16.

classes/bundle.bbclass
```diff
@@ -58,8 +58,16 @@ RAUC_BUNDLE_BUILD       ??= "${DATETIME}"
...
+        slotflags = d.getVarFlags('RAUC_SLOT_%s' % slot)
+        imgtype = slotflags.get('type') if slotflags else None
+        if imgtype == 'kernel' or imgtype == 'boot':
+            d.appendVarFlag('do_fetch', 'depends', ' ' + d.getVar('RAUC_SLOT_%s' % slot, True) + ':do_deploy')
```
mnhu: Same here

classes/bundle.bbclass
```diff
@@ -58,8 +58,16 @@ RAUC_BUNDLE_BUILD       ??= "${DATETIME}"
...
+        else:
+            d.appendVarFlag('do_fetch', 'depends', ' ' + d.getVar('RAUC_SLOT_%s' % slot, True) + ':do_image_complete')
```
mnhu: Same here

classes/bundle.bbclass
mnhu: Wouldn't it be a tiny bit nicer to use an intermediate `image` variable instead of duplicating `d.getVar('RAUC_SLOT_%s' % slot, True)` ?

classes/bundle.bbclass
ejoerns: It's to be consistent with the rest of the code ;) But we should re-change this in the entire code to match again current default.

classes/bundle.bbclass
ejoerns: Yes, might be a little nicer.

ejoerns:
I've added a patch to finally remove the True option to getVar() calls and based the dependency fix patch on it.

PR #20, opened 2017-09-22
Morty update

jremmet:
Maybe we can merge some newer commits also to morty? Unfortunately we have a older morty based BSP but want to use the meta-rauc layer.
meta-rauc master fails with "non-existent task do_image_complete".
Reverting commit 823cdd6 ("bundle.bbclass: depend on an earlier task for images") fixes this and newer patches like rauc-0.1.1 works.
Tested on morty

### DIFF
--- a/README
+++ b/README
@@ -39,7 +39,7 @@ III. Adding the RAUC update framework to your device
   V. References
 
 
-I. Adding the rauc Layer to your Build
+I. Adding the rauc Layer to Your Build
 ======================================
 
 In order to use this layer, you need to make the build system aware of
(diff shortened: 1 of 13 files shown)


ejoerns:
Thanks for reporting and testing!

I agree with you that we should fix issues in morty and backport fixes that will be compatible. What I am not sure about is if we want to have version bumps in the release branch, but maybe bugfix releases are ok.

For the `do_image_complete` issue I've a fix in #19 (that I still have to resend with my fixes... :see_no_evil: )

jremmet:
At least bundle.class fixes are an issue with using meta-rauc with morty since yocto don't allow to modify them via append.

ejoerns:
I've now pushed some non-critical fixes to morty.

About the build issue: In your series you both added and removed the critical commit "bundle.bbclass: depend on an earlier task for images", thus omitting it should be sufficient for you? Otherwise I could propose an adapted version of the upstream fix for this, but I have no BSP to test this atm.

jremmet:
omitting the patch works for me also.

ejoerns:
Great! Could you test if the current upstream changes we now have in meta-rauc 'morty' branch are sufficient for you? In this case I would close this PR.

ejoerns:
Thanks!

PR #21, opened 2017-10-02
classes: bundle: fix fetchall for bundles

hundeboll:
When running `bitbake -c fetchall some-bundle` the entire contents of
the bundle end up being build due to the dependencies assignled to
`do_fetch`. This kind of breaks the point of do_fetchall.

Instead of using `do_fetch()` to extract images from
`${DEPLOY_DIR_IMAGE}`, a new `do_stage` task is added between `do_fetch`
and `do_unpack`.

### DIFF
--- a/classes/bundle.bbclass
+++ b/classes/bundle.bbclass
@@ -38,7 +38,7 @@ PACKAGE_ARCH = "${MACHINE_ARCH}"
 
 RAUC_IMAGE_FSTYPE ??= "${@(d.getVar('IMAGE_FSTYPES') or "").split()[0]}"
 
-do_fetch[cleandirs] = "${S}"
+do_stage[cleandirs] = "${S}"
 do_patch[noexec] = "1"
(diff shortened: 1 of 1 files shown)


ejoerns:
You are totally right with the general issue, fetch should not have build-dependencies. But I would prefer a different solution for this than introduce a new `do_stage` task. This could also be misleading when referring to OE Legacy Staging that was removed from oe-core some time ago.

I've played with supporting also custom manifests files for more complex scenarios, therefor I require to have a working do_fetch stage, anyway and I moved code to do_unpack. I obviously missed also fixing the `do_fetch` dependencies and make them `do_unpack` dependencies. I will push a PR with this shortly.

mnhu:
Thanks. Looking forward to your PR, as my fix in our own layer is anything but beautiful...

ejoerns:
Here it is: #24

ejoerns:
Now fixed by #24. Thx again for pointing this out!

PR #22, opened 2017-10-24
classes/bundle: use deploy mechanism to be properly handled by sstate

ejoerns:
This will automatically clean up previously build bundle artifacts and
remain only the last one, such as it is done for all other images, etc.

This also moves the created bundles from `bundle/` subfolder to the
default image folder. Having a separate `bundle/` folder will not be
required anymore due to the reduced amount of bundles remaining.


ejoerns:
Fixed cleandir handling, added warning if existing 'old-style' `bundles/` deploy dir was detected.

PR #25, opened 2017-11-10
add basic example bundle recipe based on core-image-minimal

ejoerns:
This provides both an initial guess for creating your own while allowing
to be a reference for simple automated tests.

Signed-off-by: Enrico Joerns <ejo@pengutronix.de>

### DIFF
--- a/recipes-core/bundles/core-bundle-minimal.bb
+++ b/recipes-core/bundles/core-bundle-minimal.bb
@@ -1,0 +1,22 @@
+# A minimal demo bundle
+#
+# Note: The created bundle will not contain RAUC itself yet!
+# To add this, properly configure it for your specific system and add it to
+# your image recipe you intend to build a bundle from:
+#
+# IMAGE_INSTALL_append = " rauc"
+#
+# Also note that you need to configure RAUC_KEY_FILE and RAUC_CERT_FILE to
+# point to contain the full path to your key and cert.
+# Depending on you requirements you can either set them via global
+# configuration or from a bundle recipe bbappend.
+#
+# For testing purpose, you may use the scripts/openssl-ca.sh to create some.
+
+inherit bundle
+
+RAUC_BUNDLE_COMPATIBLE ?= "Demo Board"
+
+RAUC_BUNDLE_SLOTS ?= "rootfs"
+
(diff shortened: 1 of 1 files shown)


recipes-core/bundles/core-bundle-minimal.bb
```diff
@@ -0,0 +1,16 @@
...
+#
+# For testing purpose, you may use the scripts/openssl-ca.sh to create some.
+
+inherit bundle
+
+RAUC_BUNDLE_COMPATIBLE ?= "Demo Board"
+
+RAUC_BUNDLE_SLOTS ?= "rootfs"
+
+RAUC_SLOT_rootfs ?= "core-image-minimal"
```
jluebbe: Do we make sure that rauc is available on the target? If not, we should have a comment explaining that.

ejoerns: We could make sure RAUC is on the target, but we would not be able to ship it with a working configuration, thus this would not help much. ATM I would prefer having a comment noting that.

jluebbe: OK, sounds reasonable.

ejoerns:
Updated description:

``` diff
diff --git a/recipes-core/bundles/core-bundle-minimal.bb b/recipes-core/bundles/core-bundle-minimal.bb
index 374a029..accf183 100644
--- a/recipes-core/bundles/core-bundle-minimal.bb
+++ b/recipes-core/bundles/core-bundle-minimal.bb
@@ -1,7 +1,13 @@
 # A minimal demo bundle
 #
-# Note that you need to configure RAUC_KEY_FILE and RAUC_CERT_FILE to point to
-# contain the full path to your key and cert.
+# Note: The created bundle will not contain RAUC itself yet!
+# To add this, properly configure it for your specific system and add it to
+# your image recipe you intend to build a bundle from:
+#
+# IMAGE_INSTALL_append = " rauc"
+#
+# Also note that you need to configure RAUC_KEY_FILE and RAUC_CERT_FILE to
+# point to contain the full path to your key and cert.
 # Depending on you requirements you can either set them via global
 # configuration or from a bundle recipe bbappend.
 #
```

PR #29, opened 2017-12-14
classes: bundle: overwrite default slotname

jremmet:
If you want to create bundles for variants you need image names like
"kernel.variant,1". This doesn't work with VarFlags. Allow to overwrite
the default image name with the name flag-

Signed-off-by: Jan Remmet <j.remmet@phytec.de>


jremmet:
update commit message

PR #30, opened 2018-01-17
depend on do_deploy for 'file'-SLOTS.

leiflm:
I assume that `file` is not intended to be an image (rootfs)?

If that assumptions is correct, the provided patch should be applied.

Signed-off-by: Leif Middelschulte <leif.middelschulte@gmail.com>

### DIFF
--- a/classes/bundle.bbclass
+++ b/classes/bundle.bbclass
@@ -63,7 +63,7 @@ python __anonymous() {
         slotflags = d.getVarFlags('RAUC_SLOT_%s' % slot)
         imgtype = slotflags.get('type') if slotflags else None
         image = d.getVar('RAUC_SLOT_%s' % slot)
-        if imgtype == 'kernel' or imgtype == 'boot':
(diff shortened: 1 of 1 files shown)


ejoerns:
You are right, but I am not sure if we shouldn't do it the other way round and say something like

    if imgtype == 'image':
        d.appendVarFlag('do_unpack', 'depends', ' ' + image + ':do_image_complete')
    else:
        d.appendVarFlag('do_unpack', 'depends', ' ' + image + ':do_deploy')

What do you think?

leiflm:
> What do you think?

Sure. I just wanted my change to be as little invasive as possible here.

Do you  want me to update the PR?

ejoerns:
Yep :smile:

leiflm:
there you go. Sorry for multiple commits. Changed the commit message and sign-off 😬

ejoerns:
Thanks!

I've squashed it and adapted the git message header to be consistent with other changes to bundle.bbclass

ejoerns:
Manually cherry-picked to master

PR #31, opened 2018-01-22
Add support for building RAUC with casync support

ejoerns:
This adds basic OE integration for casync, zstd compression lib and adds the RAUC casync support patches as posted [here](https://github.com/rauc/rauc/pull/217).

### DIFF
--- a/README
+++ b/README
@@ -26,6 +26,11 @@ For rauc-hawkbit client:
   layers: meta-python
   branch: master
 
+For casync:
+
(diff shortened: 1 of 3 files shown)


ejoerns:
Updated PR to latest meta-rauc master version. This allowed to remove the additional RAUC patchstack as casync was merged to RAUC. The casync recipe I've updated to latest git master revision and clean-up both casync and zstd recipe.

PR #33, opened 2018-01-30
make the BUNDLE_NAME and the dependent variables overridable

EugenWiens:
In some use cases it is necessary to override the bundle name. Changed the value assignment to lowest priority, so that is in this use case cleaner.

Signed-off-by: Eugen Wiens <eugen.wiens@jumo.net>


ejoerns:
replaced by #35

PR #36, opened 2018-01-31
add check if imgtype is set

EugenWiens:
in the case if the image type is not set, the behavior of the dependency can be wrong. Add fatal if the type is not set to prevent unexpected behavior.

Signed-off-by: Eugen Wiens <eugen.wiens@jumo.net>


ejoerns:
Damn, seems you are right. I assumed the default to be 'image' already, but did not check it in code.

I guess to be compatible with former behavior it is best to either handle 'imgtype' `None` separately or (which i would prefer as it is more obvious) default to imgtype `'image'`

    imgtype = slotflags.get('type') if slotflags else 'image'

What do you think?

EugenWiens:
It is fine for me. Maybe an default behavior is better.

EugenWiens:
Should I update the pull request ?

ejoerns:
Feel free ;)

PR #39, opened 2018-02-08
Rocko+rauc 0.3

leiflm:
The first commit adds a recipe for [rauc v0.3](https://github.com/rauc/rauc/releases/tag/v0.3).

The second commit updates the provided `_git.bb` to point to the [0.3 revision](https://github.com/rauc/rauc/commit/419387685f1e5fb0d59aa21d0eb47ddae5120600).

Even though there is no `_0.1.bb` I left the `0_0.2.bb` there for users to set `PREFERED_VERSION_rauc = 0.2` in their `distro.conf`.

### DIFF
--- a/recipes-core/rauc/rauc-git.inc
+++ b/recipes-core/rauc/rauc-git.inc
@@ -10,3 +10,5 @@ PV = "0.3+git${SRCPV}"
 S = "${WORKDIR}/git"
 
 SRCREV = "419387685f1e5fb0d59aa21d0eb47ddae5120600"
+
+DEFAULT_PREFERENCE = "-1"
--- a/recipes-core/rauc/rauc-git.inc
+++ b/recipes-core/rauc/rauc-git.inc
@@ -1,12 +1,12 @@
 require rauc.inc
 
-PR = "r9"
+PR = "r0"
 
 SRC_URI = " \ 
   git://github.com/rauc/rauc.git;protocol=https \
   "
 
-PV = "0.1.1+git${SRCPV}"
+PV = "0.3+git${SRCPV}"
 S = "${WORKDIR}/git"
 
-SRCREV = "4b6453e8678a48aa793c41871e3bfcfdbafce881"
+SRCREV = "419387685f1e5fb0d59aa21d0eb47ddae5120600"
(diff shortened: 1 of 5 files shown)


recipes-core/rauc/rauc-git.inc
```diff
@@ -1,12 +1,12 @@
 require rauc.inc
 
-PR = "r9"
+PR = "r1"
```
ejoerns: We can even set it to `r0` as the docu suggests

recipes-core/rauc/rauc-git.inc
```diff
@@ -1,12 +1,12 @@
...
 
 SRC_URI = " \ 
   git://github.com/rauc/rauc.git;protocol=https \
   "
 
-PV = "0.1.1+git${SRCPV}"
+PV = "0.3+git${SRCPV}"
```
ejoerns: Moving this to 0.3 will mask the standard rauc_0.3.bb recipe as the version here is higher due to the appended git string.

We should set `DEFAULT_PREFERENCE` to -1 here as a workaround.

ejoerns:
I would also like to apply this to master first.

I am not sure how to deal with non-bugfix version updates in feature branches, yet. At the moment I would like to keep the rocko branch still at 0.2.

recipes-core/rauc/rauc-git.inc
leiflm: that's fine with me 👍

leiflm:
> At the moment I would like to keep the rocko branch still at 0.2

How about keeping it at `0.2` by default, but add the `0.3` recipe? I.e. using `DEFAULT_PREFERENCE=-1` on the `rauc_0.3.bb` too for rocko?

ejoerns:
Well, I would like to avoid that.
To be honest `DEFAULT_PREFERENCE` is a bad hack and there are so many ways to shoot yourself in the food with yocto package version selection (have tried it out in different ways the last days...) that I would like to prevent the confusion it might confuse having a 0.3 recipe but letting the 0.2 recipe build.
Otherwise this is a new test vector that will not be well covered and might have undesired side-effects.

For now I would add it to master and then decide if we backport it to rocko, but only the same way as it is in master.

ejoerns:
Applied to master

ejoerns:
Well, I've just discussed this with @jluebbe and we agreed on providing rauc 0.3 for rocko release, too.
I've cherry-picked and pushed all the relevant changes.

PR #40, opened 2018-02-15
Further fixes for bundle class varflag handling

ejoerns:
This adds two patches that fix / improve the varflag handling in the bundle class


ejoerns:
Updated this by reducing the level for reporting unset type to debug (note will always pollute the console).

added 'depends' varflag

PR #41, opened 2018-03-08
rauc-native_0.2: Use DEPLOY_DIR_TOOLS

krevsbech:
Applying fix from 7f5c3e024747ec32f5b61c8aea8fd5fc4da152ed to rauc_native_0.2
also.

Signed-off-by: Kasper Revsbech <krev@triax.com>


krevsbech:
Please note that the issue might also apply too rauc_native_0.3 and that the rauc_native_0.2.bb also should be fixed in branch:rocko
I can create a PR against rocko if you like or you can cherry-pick your-self if you agree that it should also be fixed in rocko.

ejoerns:
cherry-picked to rocko

Would you like to add a PR for 0.3?

krevsbech:
I'll do and test the 0.3 tomorrow.

PR #52, opened 2018-09-17
two small bundle.bbclass tweaks

Villemoes:


ejoerns:
Cherry-picked to sumo branch

PR #58, opened 2018-11-23
do not use {S}ource dir as bundle contents dir

ilkondr:
S = ${WORKDIR}/bundle can lead to errors like that:

```
WARNING: dummy-bundle-1.0-r0 do_populate_lic: Could not copy license file .../COPYING: [Errno 2] No such file or directory: '.../dummy-bundle/1.0-r0/bundle/COPYING'
ERROR: dummy-bundle-1.0-r0 do_populate_lic: QA Issue: dummy-bundle: LIC_FILES_CHKSUM points to an invalid file: .../dummy-bundle/1.0-r0/bundle/COPYING [license-checksum]
```

Since bitbake uses S variable as a path to source directory it seems correct to use separate dir for bundle contents.

### DIFF
--- a/classes/bundle.bbclass
+++ b/classes/bundle.bbclass
@@ -90,8 +90,9 @@ python __anonymous() {
             d.appendVarFlag('do_unpack', 'depends', ' ' + image + ':do_deploy')
 }
 
-S = "${WORKDIR}/bundle"
+S = "${WORKDIR}"
 B = "${WORKDIR}/build"
+BUNDLE_DIR = "${S}/bundle"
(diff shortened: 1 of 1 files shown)


ejoerns:
An alternative would be to use
`LIC_FILES_CHKSUM = "file://${WORKDIR}/COPYING;md5=xxxx`

But following the argumentation that everything (except this what is pulled in from deploy dir) that is a potential 'source' comes from the BSP itself, it would be valid to set the common `S = "${WORKDIR}"`.

PR #62, opened 2018-12-21
Support for RAUC v1.0

ejoerns:
This removes recipes for obsolete RAUC versions v0.3 and v0.4 and replaces them by the RAUC v1.0 recipe.


ejoerns:
@OnkelUlla You're right, I missed an 'is' and wrote '1.0' where '0.4' was meant. Fixed and also clarified the prior 'which' by replacing it with 'The v1.0 version'

PR #66, opened 2019-01-11
zstd: add nobranch to SRC_URI to fix fetchingA

ejoerns:
It seems that the git sha referencing the release tag 1.3.3 is not
contained in the master branch anymore and thus not found by the default
fetching magic.

We use 'nobranch=1' option to not limit sha lookup on master branch.


ejoerns:
352c3be ('zstd: add nobranch to SRC_URI to fix fetching ') manually applied to sumo (as 04d3dd6) and thud (as b6dd750) branch.

Issue #63, opened 2019-01-08
add rauc 1.0 support in sumo branch


thekyz:
As the title says, I'm a bit reluctant on using the master branch and we need 1.0 for full uboot support.

ejoerns:
@thekyz I would see the latest 2 poky release still under maintenance as potential candidates for adding 1.0 support. Thus would include sumo and thud.

It was the plan to add this, anyway. But I wanted to have support in master first before backporting it to older versions.

thekyz:
Great; do you have an ETA then?

ejoerns:
ETA is ASAP :)

Will talk to @jluebbe how it will be added.

ejoerns:
Fixed by manually PR #67 to sumo branch, too.

PR #67, opened 2019-01-13
[thud] rauc: add RAUC 1.0 recipe

ejoerns:
This adds RAUC 1.0 to the existing set of recipes.

It is highly recommended to switch to the 1.0 release that is fully
compatible with 0.4 but fixes several potential issues.
However, for those that need to stick to an older version for whatever
reason, the old recipes are kept.

You can set an explicit version as follows somewhere in your global
configuration:

    PREFERRED_VERSION_rauc = "0.4"


ejoerns:
Manually applied to 'sumo' as d6360ed.

Fixes #63

PR #69, opened 2019-01-16
busybox: add bbappend to enable FEATURE_TAR_LONG_OPTIONS

timonmentink:
Sumo branch is not working at this moment because of the busybox config FEATURE_TAR_LONG_OPTIONS.
This pullrequest will add the necessary files to the sumo branch.

"Commit message from what i copied into the sumo branch"
This wil be required by newer RAUC versions to call tar with
`--numeric-owner` in order to preserve original uid/gid during
extraction.


ejoerns:
Yep, you're absolutely right. I already wanted to do this with the RAUC 1.0 series but seems I was interrupted at the wrong time.
This patch is valid for thud and master, too (but in all cases only if you use tar archives as artifacts).

ejoerns:
Ah, I got confused. It is indeed only missing on sumo and I applied this to master/thud in August already.. Thus all is fine now. Thanks for testing and fixing!

PR #72, opened 2019-02-21
backport to rocko: bundle.bbclass: do not create bundle in content dir

Seidel-Michael:
backport to rocko:

Currently, RAUC bundles contain an empty bundle.raucb file. Presumably,
mksquashfs creates the output file before starting to walk over the
content dir, and hence it picks up the image it is creating. It doesn't
seem to hurt currently, but it's easy to imagine how things might go
wrong, and it's certainly confusing and unnecessary to have that file
inside the bundle.

The logic in bundle.bbclass is already prepared for using ${S} and ${B}
for distinct purposes, so just make sure that they are indeed distinct.

Signed-off-by: Michael Seidel <github@seidel-michael.de>

### DIFF
--- a/classes/bundle.bbclass
+++ b/classes/bundle.bbclass
@@ -73,7 +73,9 @@ python __anonymous() {
             d.appendVarFlag('do_unpack', 'depends', ' ' + image + ':do_deploy')
 }
 
-S = "${WORKDIR}/bundle"
+S = "${WORKDIR}"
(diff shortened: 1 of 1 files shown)


Seidel-Michael:
The issue in the rocko branch can be fixed by adding
`B = "${WORKDIR}/build"`
to the bundle *.bb file.

This pull request is obsolete.

PR #74, opened 2019-04-03
rauc-mark-good.service: leverage systemd's sync target

leiflm:
systemd has a dedicated, special target called `boot-complete.target` that indicates,
that a boot-up is considered successful.

Citing the documentation[0]:
> This target is intended as generic synchronization point for services that shall determine or act on whether the boot process completed successfully. [...]

[0] https://www.freedesktop.org/software/systemd/man/systemd.special.html#boot-complete.target

### DIFF
--- a/recipes-core/rauc/files/rauc-mark-good.service
+++ b/recipes-core/rauc/files/rauc-mark-good.service
@@ -1,5 +1,7 @@
 [Unit]
 Description=Rauc Good-marking Service
+After=boot-complete.target
+Requires=boot-complete.target
 
 [Service]
 ExecStart=@BINDIR@/rauc status mark-good


recipes-core/rauc/files/rauc-mark-good.service
```diff
@@ -1,5 +1,7 @@
 [Unit]
 Description=Rauc Good-marking Service
+After=local-fs.target boot-complete.target
```
OnkelUlla: At least in systemd v241 the _boot-complete.target_ has directives `Requires=sysinit.target` and `After=sysinit.target` with _sysinit.target_ having directives `Wants=local-fs.target` and `After=local-fs.target`. So isn't `After=boot-complete.target` enough with the additional `After=local-fs.target` directive being redundant here as it is implied by the former?

leiflm: You might be right. 

I just mimiced systemd's own [`systemd-bless.boot.service`](https://github.com/systemd/systemd/blob/master/units/systemd-bless-boot.service.in) in this regard.

Further information on can be found in [systemd's documentation](https://github.com/systemd/systemd/blob/master/docs/AUTOMATIC_BOOT_ASSESSMENT.md).

leiflm: @OnkelUlla so what's the preference here?
- implicit directives?
- mimicking `systemd-bless.boot.service`

OnkelUlla: Hi Leif, starting with our discussion at last year's ELCE I am still wearing my wristwatch on the "wrong" side to remind me of this pull request, so it is finally time to change that! ;-)

Other than `rauc-mark-good.service` which indirectly acts upon a device file `systemd-bless.boot.service` needs one particular local filesystems to do its work. I guess that's why the systemd developers explicitely added the `After=local-fs.target` directive. So in our case I still vote for removing that unneded explicit ordering.

leiflm: > it is finally time to change that!

Glad you found a moment :)

> Other than `rauc-mark-good.service` which indirectly acts upon a device file

Hm... it seems that a proper fix depends on the backend. Doesn't it?
I.e. the barebox backend [needs the `barebox-state`](https://github.com/rauc/rauc/blob/98d1416519037570ffd82352999ca044fa81e68c/src/bootchooser.c#L162) executable mounted (and executable). The tool, in turn, needs the state device to be available.

So, an exemplary systemd notation that would be like:
```ini
[Unit]
After=dev-mmcblk3.device # in case that's the state device
Requires=dev-mmcblk3.device
RequiresMountsFor=/usr/sbin/barebox-state
ConditionFileIsExecutable=/usr/sbin/barebox-state
```

The U-Boot backend would likely need modifications.

OnkelUlla: > Hm... it seems that a proper fix depends on the backend. Doesn't it?

Right, the requirements depend on the backend. But from my perspective there is nothing to _fix_ in meta-rauc's general service unit file at the moment.

> I.e. the barebox backend [needs the `barebox-state`](https://github.com/rauc/rauc/blob/98d1416519037570ffd82352999ca044fa81e68c/src/bootchooser.c#L162) executable mounted (and executable). The tool, in turn, needs the state device to be available.
>
> So, an exemplary systemd notation that would be like:
>
> ```ini
> [Unit]
> After=dev-mmcblk3.device # in case that's the state device
> Requires=dev-mmcblk3.device
> ```

If you really need something like that in an individual application of RAUC just add it via a corresponding configuration file in the `rauc-mark-good.service.d` drop-in directory from within your meta-bsp layer.

> ```ini
> RequiresMountsFor=/usr/sbin/barebox-state
> ```

Our `rauc-mark-good.service` already has an implicit `After=local-fs.target` as it implicitely [requires to run after `sysinit.target`](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Default%20Dependencies). So for me this looks much too fine grained.

> ```ini
> ConditionFileIsExecutable=/usr/sbin/barebox-state
> ```

This is something that we normally do not check in a unit file, but fix in the build system if it's broken. Again: if it is something that you need in your individual situation add it via a drop-in configuration file.

So let's come back to your initial intent of having `rauc-mark-good.service` _require_ to be run only _after_ `boot-complete.target` which I totally agree is a good thing to have!

OnkelUlla: > Hm... it seems that a proper fix depends on the backend. Doesn't it?

Right, the requirements depend on the backend. But from my perspective there is nothing to _fix_ in meta-rauc's general service unit file at the moment.

leiflm: > just add it via a corresponding configuration file in the rauc-mark-good.service.d drop-in directory from within your meta-bsp layer.

Agreed.

> So let's come back to your initial intent of having `rauc-mark-good.service` require to be run only after `boot-complete.target` which I totally agree is a good thing to have!

Right :) I've rebased and update the commit accordingly.

ejoerns:
cherry-picked to zeus branch (https://github.com/rauc/meta-rauc/commit/ff928cdf6aaec7c13c75c4f0acdc72659782926b)

PR #77, opened 2019-06-11
classes: bundle: support [file] varflag for type 'image'

danielbatw:
Allow to use a non-autogenerated image name.

Signed-off-by: Gero Schumacher <gero.schumacher@peiker-cee.de>

### DIFF
--- a/classes/bundle.bbclass
+++ b/classes/bundle.bbclass
@@ -142,7 +142,10 @@ def write_manifest(d):
             img_fstype = slotflags.get('fstype')
 
         if imgtype == 'image':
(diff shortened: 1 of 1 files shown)


ejoerns:
@danielbatw is this a backport request? Because the patch is in *master* and *warrior* branch already (since version https://github.com/rauc/meta-rauc/pull/70).

By default, we only support the two most recent releases (*thud* and *warrior* as of now). Are you able to move to a recent yocto version?

danielbatw:
Yes it is. I can't move to thud or warrior at the moment, so that would be helpful. Feel free to ignore it, I can live without it.

danielbatw:
I'm closing this since I need more commits backported.

PR #81, opened 2019-07-02
Update of casync package

ejoerns:
* update to latest master (no release yet)
* add nativesdk build


ejoerns:
I've pushed an update to fix udev handling when building for nativesdk.

ejoerns:
cherry-picked to thud branch as it was not compiling on this either.

PR #82, opened 2019-07-05
Add CI testing via circleci

ejoerns:
circleci allows us to cache results which enables saving and restoring sstate-cache to speed up builds.

This, combined with using sstate.yoctoproject.org reduces the build time of meta-rauc components to a few minutes and makes it appropriate for CI testing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,36 @@
+version: 2.1
+
+jobs:
+  build:
+    docker:
+      - image: crops/yocto:ubuntu-18.10-builder
+        environment:
+          LC_ALL: en_US.UTF-8
+          LANG: en_US.UTF-8
+          LANGUAGE: en_US.UTF-8
+    steps:
+      - checkout
+      - run:
+          name: Clone poky
+          command: git clone git://git.yoctoproject.org/poky
+      - run:
+          name: Clone meta-oe
+          command: git clone https://github.com/openembedded/meta-openembedded.git meta-oe
+      - run:
+          name: Link meta-rauc
+          command: ln -sf . meta-rauc
+      - run:
+          name: Create meta-this
+          command: mkdir -p meta-this/conf/ && cp .circleci/*.sample meta-this/conf/.
+      - run:
+          name: Initialize build directory
+          command: TEMPLATECONF=../meta-this/conf source poky/oe-init-build-env build
+      - run:
+          name: Run bitbake
+          command: source poky/oe-init-build-env build && bitbake rauc
+
+
+workflows:
+  workflow:
+    jobs:
+      - build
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,50 @@
...
+          command: ln -sf . meta-rauc
+      - run:
+          name: Create meta-this
+          command: mkdir -p meta-this/conf/ && cp .circleci/*.sample meta-this/conf/.
+      - run:
+          name: Source env (changes into 'build')
+          command: TEMPLATECONF=../meta-this/conf source poky/oe-init-build-env build
+      - run:
+          name: Properly set LANG to UTF-8-supported
+          command: printf "export LC_ALL=en_US.UTF-8\nexport LANG=en_US.UTF-8\nexport LANGUAGE=en_US.UTF-8" >> $BASH_ENV
```
jluebbe: Could we use explicit environment setup instead?
https://circleci.com/docs/2.0/env-vars/#setting-an-environment-variable-in-a-job

.circleci/config.yml
```diff
@@ -0,0 +1,50 @@
...
+          command: git clone https://github.com/openembedded/meta-openembedded.git meta-oe
+      - run:
+          name: Link meta-rauc
+          command: ln -sf . meta-rauc
+      - run:
+          name: Create meta-this
+          command: mkdir -p meta-this/conf/ && cp .circleci/*.sample meta-this/conf/.
+      - run:
+          name: Source env (changes into 'build')
+          command: TEMPLATECONF=../meta-this/conf source poky/oe-init-build-env build
```
jluebbe: Shouldn't this be named "Initialize build directory"? The following steps all source poky/oe-init-build-env as well, so there's nothing kept from step to step.

.circleci/config.yml
```diff
@@ -0,0 +1,50 @@
...
+  build:
+    docker:
+      - image: crops/yocto:ubuntu-18.10-builder
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - sstate-test-{{ .Branch }}-{{ .Revision }}
+            - sstate-test-{{ .Branch }}-
+            - sstate-test-
```
jluebbe: You could add a sstate-test-master before the general fallback.

jluebbe:
I've enabled meta-rauc on circle.ci: https://circleci.com/gh/rauc/meta-rauc

.circleci/config.yml
ejoerns: circleci describes multiple ways of setting up env, but I've changed to use the yaml config way now

.circleci/config.yml
ejoerns: Yes. Fixed.

.circleci/config.yml
ejoerns: Yes. Did it.

PR #91, opened 2019-10-25
rauc: rauc-target.inc: fix do_install dependency

medude:
Without the fix bitbake didn't notice that the RAUC_KEYRING_FILE was
changed automatically.

Signed-off-by: Marco Felsch <m.felsch@pengutronix.de>

### DIFF
--- a/recipes-core/rauc/rauc-target.inc
+++ b/recipes-core/rauc/rauc-target.inc
@@ -43,6 +43,8 @@ do_install () {
         install -m 755 "${WORKDIR}/rauc-mark-good.init" "${D}${sysconfdir}/init.d/rauc-mark-good"
 }
 
+do_install[vardeps] = "RAUC_KEYRING_FILE"
+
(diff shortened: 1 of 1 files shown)


ejoerns:
I cannot reproduce the behavior described. Does this happen when you change the files path in the recipe or when the content of the file changes?

Also, `vardeps` should not have any impact on the actual handling of the files content if I'm not totally wrong..

medude:
Hi Enrico,

it happened after I changed the files path therefore I used vardeps.

medude:
I tested the it again and vardeps also tracks the file content.

ejoerns:
@medude vardep will not track contents. My assumption was that it worked before and vardep did not change anything. I think this is what we verified yesterday, right?

medude:
@ejoerns you're right. I don't know why it didn't worked the first time but it works now so we can close this PR

ejoerns:
Fine :+1:

PR #92, opened 2019-10-31
rauc: update to v1.2

ejoerns:
RAUC 1.2 requires dbus pkg-config files in case of service being
enabled.


jluebbe:
We should make the dependency optional if the path is given explicitly in rauc. @michaelolbrich asked for the same already.

ejoerns:
> We should make the dependency optional if the path is given explicitly in rauc. @michaelolbrich asked for the same already.

Yes, but this is not to be solved here. Right?

PR #94, opened 2019-11-06
rauc: update to v1.2

ChHapp:
RAUC 1.2 requires dbus pkg-config files in case of service being
enabled.


ChHapp:
This PR got obsoleted so I close it now.

PR #95, opened 2019-11-06
rauc-target: disable the watching of a block device node with inotify

bith3ad:
Barebox-state, mbr-switch and eMMC-bootloader updates operates on the
root device node. This triggers the inotify which causes a partition
rescan. This is racy as devices might disappear while RAUC tries to
access them during a update.

Signed-off-by: Marco Felsch <m.felsch@pengutronix.de>

### DIFF
--- a/recipes-core/rauc/files/61-block-nowatch.rules
+++ b/recipes-core/rauc/files/61-block-nowatch.rules
@@ -1,0 +1,5 @@
+# Don't watch metadata changes, caused by tools closing the device node which was opened for writing.
+# Barebox-state, mbr-switch and eMMC-bootloader update operates on the root device node. This triggers
+# the inotify which causes a partition rescan. This is racy as devices might disappear while RAUC
+# tries to access them during a update.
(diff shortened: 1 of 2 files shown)


ejoerns:
Please rebase onto current master to enable CI testing.

ejoerns:
However, while the watch in our setup mainly affects rauc, I would not see this as belonging to the rauc package. It is more of general system handling. Thus please consider adapting the recipe that originally brought in the other udev rule we try to patch here. Or create a custom one.

medude:
@ejoerns hm it's up to you. We can add a .bbappend for systemd to override the default but then it gets overridden anytime. Should that be the case if we include the meta-rauc into our bsp?

ejoerns:
No it should not be the default when adding. But this is a question on a different level that I want to target within the next days. The general idea is to make rauc a DISTRO_FEATURE and then enable all the more or less special stuff only if this feature is explicitly selected.
@michaelolbrich @jluebbe any opinions on the scope of the rules file?

ejoerns:
Is this still relevant? Or should we close it?

PR #99, opened 2019-11-15
zstd: remove recipe from layer in favor of its meta-oe version

ejoerns:
Starting with the zeus release, meta-oe contains a recipe for zstd which
is more recent (1.4.3) than the one we have in meta-rauc (1.4.0).
We already depend on meta-oe, thus we can safely remove this recipe here.


jluebbe:
The build failed because meta-oe/recipes-extended (where the zstd recipe lives) is not in the CI's layer configuration.

ejoerns:
Working with experts :laughing:  They removed zstd from meta-oe master but it is not merge in poky master, yet...

PR #101, opened 2019-11-21
Minor fixes, cleanup and documentation for bundle.bbclass

ejoerns:
This small series:

* fixes variable names in python code
* adds variable documentation varflags
* adds note printouts for adding artifacts to bundle dir
* removed old-style bundle layout warning

### DIFF
--- a/classes/bundle.bbclass
+++ b/classes/bundle.bbclass
@@ -133,10 +133,10 @@ def write_manifest(d):
     for slot in (d.getVar('RAUC_BUNDLE_SLOTS') or "").split():
         slotflags = d.getVarFlags('RAUC_SLOT_%s' % slot)
         if slotflags and 'name' in slotflags:
-            imgname = slotflags.get('name')
+            slotname = slotflags.get('name')
         else:
-            imgname = slot
-        manifest.write('[image.%s]\n' % imgname)
+            slotname = slot
+        manifest.write('[image.%s]\n' % slotname)
         if slotflags and 'type' in slotflags:
             imgtype = slotflags.get('type')
         else:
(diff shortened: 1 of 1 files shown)


classes/bundle.bbclass
```diff
@@ -133,10 +133,10 @@ def write_manifest(d):
...
         if slotflags and 'name' in slotflags:
-            imgname = slotflags.get('name')
+            slotname = slotflags.get('name')
         else:
-            imgname = slot
-        manifest.write('[image.%s]\n' % imgname)
+            slotname = slot
+        manifest.write('[image.%s]\n' % slotname)
         if slotflags and 'type' in slotflags:
             imgtype = slotflags.get('type')
```
OnkelUlla: What about changing `imgtype` to `slottype` here and in line 74 and the following as well? And then there are `img_fstype` in line 146 and following and `imgsource` in line 150 and following, too.

ejoerns: In contrast to (now) 'slotname', the variables you mention actually describe properties of the image. `imgtype` is describes what kind of image this (more from the build systems perspective) is but has nothing to do with the RAUC understanding of 'image type' or 'slot type. `img_fstype` in turn is used to guess the right file name extension of the the *image* to fetch. Thus also unrelated to slots, `imgsource` is the path where we finally fetch the image from.

On the other hand: 'imgtype', the one that I renamed, was used with the same name for very different purpose: The first one was for the targeted slot class name (the one now renamed) the second was for the already-mentioned type of build system image to use.

ejoerns:
Fixed the commit typo. Thanks!

classes/bundle.bbclass
OnkelUlla: Okay, just wanted to be sure that you already though about it: your reasoning is plausible - in fact there is some ambiguity here.
So: go for it! :-)

PR #102, opened 2019-11-21
classes: bundle: add simple mechanism for adding extra artifacts to bundle

### LINKED ISSUE #17: Include arbitrary files in the bundle
In a custom post-install hook embedded in a rauc bundle, I'm trying to install specific files that require specific treatment (in this specific case, the bootloader itself).

I initially tried to declare it as a "virtual" slot in ``RAUC_BUNDLE_SLOTS`` of type ``file``. I was successfully able to package an arbitrary file from the deploy dir as "somefile.img" at the root of the bundle. This also create the corresponding declaration in the manifest.

Unfortunately, the current live system does not have the corresponding virtual slot declaration in its system.conf file and fails to install the bundle, which is expected.

I solved this with a do_fetch_append (see below) In the bundle recipe, with the related code in the main slot post-install hook:

```
python do_fetch_append() {
    bootloader_name = 'bootloader'
    deploy_dir = d.expand("${DEPLOY_DIR_IMAGE}")
    bundle_dir = d.expand("${S}/bundle")

    shutil.copytree(
        os.path.join(deploy_dir, bootloader_name),
        os.path.join(bundle_dir, bootloader_name),
    )
}
```

Maybe we could design a better way to embed any arbitrary files/folder inside a rauc bundle using meta-rauc. It could look like:

```
RAUC_BUNDLE_FILES = "file1 file2 ... fileN"
```

Which would copy files from the deploy directory to the bundle.

If this looks good to you, I can work on this.

ejoerns:
This introduces variables `RAUC_BUNDLE_EXTRA_DEPENDS` and
`RAUC_BUNDLE_EXTRA_FILES`.

With RAUC_BUNDLE_EXTRA_FILES one specifies a space-separated list of
artifacts that are first search in DEPLOY_DIR_IMAGE and then in WORKDIR.

This allows both to add artifact via SRC_URI as well as by grabbing the
generated files from other deploy recipes.

To ensure the artifacts are placed in the deploy dir before fetching
them bundle generation one must add their recipe's name to the
RAUC_BUNDLE_EXTRA_DEPENDS variable:

    RAUC_BUNDLE_EXTRA_DEPENDS += "recipe-name"

The dependency defaults to the do_deploy task. If something else is
required, than one can optionally do this with the recipe-deptask
notation:

    RAUC_BUNDLE_EXTRA_DEPENDS += "recipe-name:do_<taskname>"

Fixes #17


classes/bundle.bbclass
```diff
@@ -33,6 +33,18 @@
...
+# To add additional artifacts to the bundle you can use RAUC_BUNDLE_EXTRA_FILES
+# and RAUC_BUNDLE_EXTRA_DEPENDS.
+# For files from the WORKDIR (fetched using SRC_URI) you can write:
+#
+#   SRC_URI += "file://myfile"
+#   RAUC_BUNDLE_EXTRA_FILES += "myfile"
+#
+# For files from the DEPLOY_DIR_IMAGE (generated by another recipe) you can write:
+#
+#   RAUC_BUNDLE_EXTRA_DEPENDS += "myfile-recipe-pn"
```
ilkondr: consider adding to manual ability to provide specific task

ejoerns: You mean documenting the `my-recipe:do_<taskname>` feature? I documented this in the `[doc]' varflag. Do you think it should be noted here, too?

ilkondr: Oh, now I see.
I think having it in [doc] flag is enough

ejoerns: Ok. Thanks for feedback! :+1:

ejoerns:
Missed a `bb.warn()` thath should have been a `bb.note()` -> fixed.

PR #108, opened 2020-01-22
rauc-hawkbit: Add systemd service

mschwan-phytec:
This adds a systemd service for the rauc-hawkbit client to easily start and stop it. ~By default, this service is disabled, as I think not everyone wants to use the rauc-hawkbit client on their system.~

Additionally, for an easier configuration of the client the config file is being linked to /etc.

These changes should be applied to sumo and later.

### DIFF
--- a/recipes-support/rauc-hawkbit/files/rauc-hawkbit.service
+++ b/recipes-support/rauc-hawkbit/files/rauc-hawkbit.service
@@ -1,0 +1,9 @@
+[Unit]
+Description=rauc-hawkbit client service
+Requires=rauc.service
+
+[Service]
+ExecStart=@BINDIR@/rauc-hawkbit-client -c @SYSCONFDIR@/rauc-hawkbit/config.cfg
+
+[Install]
+WantedBy=multi-user.target
--- a/recipes-support/rauc-hawkbit/rauc-hawkbit_git.bb
+++ b/recipes-support/rauc-hawkbit/rauc-hawkbit_git.bb
@@ -5,13 +5,34 @@ SUMMARY = "hawkBit client for RAUC"
 
 DEPENDS = "python3-setuptools-scm-native"
 
-SRC_URI = "git://github.com/rauc/rauc-hawkbit.git;protocol=https"
+SRC_URI = " \
+    git://github.com/rauc/rauc-hawkbit.git;protocol=https \
+    file://rauc-hawkbit.service \
+"
 
 PV = "0.1.0-9+git${SRCPV}"
 SRCREV = "6a1e79b0b7c83da866e0602201db49d6e74d7dd2"
 
 S = "${WORKDIR}/git"
 
-inherit setuptools3
+inherit setuptools3 systemd
+
+PACKAGES =+ "${PN}-service"
+SYSTEMD_SERVICE_${PN}-service = "rauc-hawkbit.service"
+SYSTEMD_PACKAGES = "${PN}-service"
+
+do_install_append() {
+	install -d ${D}${sysconfdir}/${BPN}/
+	install -m 0644 ${S}/rauc_hawkbit/config.cfg ${D}${sysconfdir}/${BPN}/config.cfg
+	install -d ${D}${systemd_unitdir}/system/
+	install -m 0644 ${WORKDIR}/rauc-hawkbit.service ${D}${systemd_unitdir}/system/
+	sed -i -e 's!@BINDIR@!${bindir}!g' -e 's!@SYSCONFDIR@!${sysconfdir}!g' \
+		${D}${systemd_unitdir}/system/rauc-hawkbit.service
+}
 
 RDEPENDS_${PN} += "python3-aiohttp python3-gbulb"
+FILES_${PN}-service = " \
+    ${bindir}/rauc-hawkbit-client \
+    ${sysconfdir}/${BPN}/config.cfg \
+    ${systemd_unitdir}/system/rauc-hawkbit.service \
+"


mschwan-phytec:
I changed the title to reflect that this is work-in-progress. As the original Python implementation is merely meant to be an example and not really intended for production use, I will rework these commits to use the C implementation (https://github.com/rauc/rauc-hawkbit-updater) instead.

ejoerns:
@mschwan-phytec Having a service should be valid for both rauc-hawkbit and rauc-hawkbit-updater. We could simply leave this patch as it is and do a separate one for rauc-hawkbit-updater. Also note that I did not have time to add the rauc-hawkbit-updater recipe, yet. Thus, if you do both at once, fine :smirk:. Otherwise I'd do this within the next days.

mschwan-phytec:
> @mschwan-phytec Having a service should be valid for both rauc-hawkbit and rauc-hawkbit-updater. We could simply leave this patch as it is and do a separate one for rauc-hawkbit-updater. Also note that I did not have time to add the rauc-hawkbit-updater recipe, yet. Thus, if you do both at once, fine :smirk:. Otherwise I'd do this within the next days.

I am fine with this. I already created the recipe and will send a pull-request soon.

recipes-support/rauc-hawkbit/rauc-hawkbit_git.bb
```diff
@@ -5,13 +5,30 @@ SUMMARY = "hawkBit client for RAUC"
...
+"
 
 PV = "0.1.0-9+git${SRCPV}"
 SRCREV = "6a1e79b0b7c83da866e0602201db49d6e74d7dd2"
 
 S = "${WORKDIR}/git"
 
-inherit setuptools3
+SYSTEMD_SERVICE_${PN} = "rauc-hawkbit.service"
+SYSTEMD_AUTO_ENABLE_${PN} ?= "disable"
```
mschwan-phytec: What are the opinions on disabling the service by default? I am not sure at the moment, if it is necessary to disable it. If one wants to not use this program then not building it is the simplest option, I guess. The same applies the other way around: Using rauc-hawkbit, we want to have an enabled systemd service.

ejoerns: I'd say the default is to enable it. If one does not want rauc-hawkbit at all, one should not add it to the image. This is how almost all other packages work, too.
It might be an option to split the library part from the runtime part here as you can use rauc-hawkbit both as a simple library for integrating it in your custom python application as well as a standalone service. Adding a `rauc-hawkbit-service` runtime package that contains both the executable as well as the systemd service might be the solution.

mschwan-phytec: The systemd service is now enabled by default. I separated the systemd service to its own package, too.

recipes-support/rauc-hawkbit/files/rauc-hawkbit.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=rauc-hawkbit client service
+Requires=rauc.service
+
+[Service]
+ExecStart=/usr/bin/rauc-hawkbit-client -c /etc/rauc-hawkbit/config.cfg
```
ejoerns, line 6: To ensure flexibility, this should be

```suggestion
ExecStart=@BINDIR@/rauc-hawkbit-client -c @SYSCONFDIR@/rauc-hawkbit/config.cfg
```

with 

``` sh
sed -i -e 's!@BINDIR@!${bindir}!g' \
       -e 's!@SYSCONFDIR@!${sysconfdir}!g' ${D}${systemd_unitdir}/system/rauc-hawkbit.service
```

in do_install().

recipes-support/rauc-hawkbit/rauc-hawkbit_git.bb
```diff
@@ -5,13 +5,33 @@ SUMMARY = "hawkBit client for RAUC"
...
 
 PV = "0.1.0-9+git${SRCPV}"
 SRCREV = "6a1e79b0b7c83da866e0602201db49d6e74d7dd2"
 
 S = "${WORKDIR}/git"
 
-inherit setuptools3
+inherit setuptools3 systemd
+
+PACKAGES =+ "${PN}-service"
```
ejoerns, line 20: instead of splitting this up, consider using `SYSTEM_AUTO_ENABLE = "disable"` to not change existing behavior.

recipes-support/rauc-hawkbit/rauc-hawkbit_git.bb
```diff
@@ -14,4 +14,12 @@ S = "${WORKDIR}/git"
 
 inherit setuptools3
 
+do_install_append() {
+	install -d ${D}${sysconfdir}/${PN}/
+	ln -s ${libdir}/${PYTHON_DIR}/site-packages/rauc_hawkbit/config.cfg ${D}${sysconfdir}/${PN}/config.cfg
```
ejoerns, line 26: This makes it quite hard to place a custom config.cfg during integration here.

I'd tend to do simply install the default config here first of all and maybe later add an option to provide a custom one.

recipes-support/rauc-hawkbit/rauc-hawkbit_git.bb
```diff
@@ -14,4 +14,12 @@ S = "${WORKDIR}/git"
...
+}
+
 RDEPENDS_${PN} += "python3-aiohttp python3-gbulb"
+FILES_${PN} += " \
+    ${sysconfdir}/${PN}/config.cfg \
```
ejoerns, line 33: I'd prefer to use the plain package name here (and in the respective install command) or at least ${BPN} here which will ensure to always be the base name, without any pre- or suffixes.

recipes-support/rauc-hawkbit/rauc-hawkbit_git.bb
mschwan-phytec: I do not quite understand. There was no systemd service before, so there is no behavior that could be changed? At the top of this PR you also mentioned that the default should be to install the service:
> I'd say the default is to enable it. If one does not want rauc-hawkbit at all, one should not add it to the image. This is how almost all other packages work, too.
> It might be an option to split the library part from the runtime part here as you can use rauc-hawkbit both as a simple library for integrating it in your custom python application as well as a standalone service. Adding a `rauc-hawkbit-service` runtime package that contains both the executable as well as the systemd service might be the solution.

recipes-support/rauc-hawkbit/files/rauc-hawkbit.service
mschwan-phytec: Done.

recipes-support/rauc-hawkbit/rauc-hawkbit_git.bb
ejoerns: Legacy me... Ok, I got the motivation for the service wrong and did not have the library aspect in mind anymore. Probably what I said initially makes more sense. :smirk: 

But legacy me also said

> Adding a rauc-hawkbit-service runtime package that contains both the executable as well as the systemd service might be the solution.

That's what confused me probably ;) The idea to split this up between the library and the application should not be limited to the service file. I'd say service, binary and configuration file should then be part of the service package. i.e., all that one does not want when using rauc-hawkbit updater as a python module.

mschwan-phytec: Ok, now I understand. Changes are pushed.

recipes-support/rauc-hawkbit/rauc-hawkbit_git.bb
mschwan-phytec: Done. The config is now installed as a normal file to `${sysconfdir}` directly instead of using a link.

recipes-support/rauc-hawkbit/rauc-hawkbit_git.bb
mschwan-phytec: Done.

ejoerns (approved):
Looks fine for me now! Thanks for the work and sorry for letting you wait.

PR #111, opened 2020-02-13
Add BUNDLE_ARGS variable.

Psykar:
While it will work as is, it probably shouldn't land until https://github.com/rauc/rauc/pull/537 lands.

### DIFF
--- a/classes/bundle.bbclass
+++ b/classes/bundle.bbclass
@@ -45,6 +45,9 @@
 #   RAUC_BUNDLE_EXTRA_DEPENDS += "myfile-recipe-pn"
 #   RAUC_BUNDLE_EXTRA_FILES += "myfile.img"
 #
+# Extra arguments may be passed to the bundle command with BUNDLE_ARGS eg:
+#   BUNDLE_ARGS += ' --mksquashfs-args="-comp zstd -Xcompression-level 22" '
+#
 # Additionally you need to provide a certificate and a key file
 #
 #   RAUC_KEY_FILE ?= "development-1.key.pem"
@@ -130,6 +133,9 @@ RAUC_KEY_FILE ??= ""
 RAUC_KEY_FILE[doc] = "Specifies the path to the RAUC key file used for signing. Use COREBASE to reference files located in any shared BSP folder."
 RAUC_CERT_FILE ??= ""
 RAUC_CERT_FILE[doc] = "Specifies the path to the RAUC cert file used for signing. Use COREBASE to reference files located in any shared BSP folder."
+BUNDLE_ARGS ??= ""
+BUNDLE_ARGS[doc] = "Specifies any extra arguments to pass to the rauc bundle command."
+
 
 DEPENDS = "rauc-native squashfs-tools-native"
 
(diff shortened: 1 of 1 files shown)


Psykar:
This should be ready to land.

classes/bundle.bbclass
```diff
@@ -279,6 +285,7 @@ do_bundle() {
 		--debug \
 		--cert="${RAUC_CERT_FILE}" \
 		--key="${RAUC_KEY_FILE}" \
+		${BUNDLE_ARGS} \
```
OnkelUlla: Use `--mksquashfs-args=$(RAUC_BUNDLE_MKSQUASHFS_ARGS)` instead here.

classes/bundle.bbclass
```diff
@@ -45,6 +45,9 @@
 #   RAUC_BUNDLE_EXTRA_DEPENDS += "myfile-recipe-pn"
 #   RAUC_BUNDLE_EXTRA_FILES += "myfile.img"
 #
+# Extra arguments may be passed to the bundle command with BUNDLE_ARGS eg:
+#   BUNDLE_ARGS += ' --mksquashfs-args="-comp zstd -Xcompression-level 22" '
```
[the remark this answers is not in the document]
ejoerns: I'd go for changing the documentation and using `BUNDLE_ARGS`. We already have the option `--signing-keyring` and I'm pretty sure there are more to be expected over time. Thus I would like to avoid adding a variable for each new option.

Psykar:
I've updated the doco to be more generic assuming that's the consensus :)

classes/bundle.bbclass
Psykar: Assuming we keep the more generic option, I'll resolve this.

classes/bundle.bbclass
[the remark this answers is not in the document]
Psykar: The other examples in the comments here aren't particuarly generic - I'm not quite sure what else to put here, as anything I put here won't be generic. Thoughts?

classes/bundle.bbclass
```diff
@@ -130,6 +133,9 @@ RAUC_KEY_FILE ??= ""
 RAUC_KEY_FILE[doc] = "Specifies the path to the RAUC key file used for signing. Use COREBASE to reference files located in any shared BSP folder."
 RAUC_CERT_FILE ??= ""
 RAUC_CERT_FILE[doc] = "Specifies the path to the RAUC cert file used for signing. Use COREBASE to reference files located in any shared BSP folder."
+BUNDLE_ARGS ??= ""
+BUNDLE_ARGS[doc] = "Specifies any extra arguments to pass the rauc bundle command."
```
OnkelUlla: Please re-add the lost "to" as in `s/to pass the rauc bundle command/to pass to the rauc bundle command/`.

Psykar: Ha, thanks.

Psykar:
The CI failure appears to be not due to my change?

ejoerns:
> The CI failure appears to be not due to my change?

Nope, it just shows that building yocto in CI with limited time and storage resources is quite painful..

PR #112, opened 2020-02-25
classes: bundle: remove executable bit from the output bundle

jluebbe:


ejoerns:
Re-triggering the test seems to have fixed the status now. circle-ci reported HTTP 500 before o.O

PR #114, opened 2020-02-27
rauc-hawkbit_git.bb add missing runtime dependency

zpfvo:
Adds missing python3-netserver dependency. Without i got the following error:

```
Traceback (most recent call last):
  File "/usr/bin/rauc-hawkbit-client", line 5, in <module>
    import aiohttp
  File "/usr/lib/python3.7/site-packages/aiohttp/__init__.py", line 6, in <module>
    from .client import (
  File "/usr/lib/python3.7/site-packages/aiohttp/client.py", line 32, in <module>
    from . import hdrs, http, payload
  File "/usr/lib/python3.7/site-packages/aiohttp/http.py", line 1, in <module>
    import http.server
  File "/usr/lib/python3.7/http/server.py", line 102, in <module>
    import socketserver
ModuleNotFoundError: No module named 'socketserver'
```

### DIFF
--- a/recipes-support/rauc-hawkbit/rauc-hawkbit_git.bb
+++ b/recipes-support/rauc-hawkbit/rauc-hawkbit_git.bb
@@ -14,4 +14,4 @@ S = "${WORKDIR}/git"
 
 inherit setuptools3
 
-RDEPENDS_${PN} += "python3-aiohttp python3-gbulb"
+RDEPENDS_${PN} += "python3-aiohttp python3-gbulb python3-netserver"


ejoerns:
@zpfvo Good catch! I ran into the same issue when updating a demo BSP based on the recent zeus branch.

This requires to be fixed upstream as it is a missing dependency for aiohttp, not for rauc-hawkbit. However, to have a temporary fix here, one would need to .bbappend the aiohttp recipe instead of the hawkit one.

I also had to add `python3-http`, but maybe this depends on the actual version one tests against.

@zpfvo It would also be great if you could add your SIgned-off-by to the commit and reference the poky / oe commit that broke things.

ejoerns:
Closed due to inactivity and as zeus is EOL for some time now, anyway.

PR #115, opened 2020-03-09
Add recipe for rauc-hawkbit-updater

mschwan-phytec:
rauc-hawkbit-updater is a simple command line tool (or daemon) written
in C (glib). It is a port of the RAUC hawkBit client written in Python.
The daemon operates as an interface between the RAUC D-Bus API and the
hawkBit DDI API.

Please apply to sumo and later.

### DIFF
--- a/recipes-support/rauc-hawkbit-updater/rauc-hawkbit-updater_git.bb
+++ b/recipes-support/rauc-hawkbit-updater/rauc-hawkbit-updater_git.bb
@@ -1,0 +1,26 @@
+SUMMARY = "The RAUC hawkBit updater operates as an interface between the RAUC D-Bus API and the hawkBit DDI API."
+HOMEPAGE = "https://github.com/rauc/rauc-hawkbit-updater"
+
+LICENSE = "LGPLv2.1"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=1a6d268fd218675ffea8be556788b780"
+
+SRC_URI = "git://github.com/rauc/rauc-hawkbit-updater.git;protocol=https"
+SRCREV = "d909982e9e4b84cb76b98bf85f25a0a88472301a"
+S = "${WORKDIR}/git"
+PV = "0.0+git${SRCPV}"
+
+inherit cmake pkgconfig systemd
+
+SYSTEMD_SERVICE_${PN} = "${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'rauc-hawkbit-updater.service', '', d)}"
+
+PACKAGECONFIG ??= " \
+    ${@bb.utils.filter('DISTRO_FEATURES', 'systemd', d)} \
+"
+PACKAGECONFIG[systemd] = "-DWITH_SYSTEMD=ON,,systemd"
+
+DEPENDS = "curl glib-2.0-native json-glib"
+
+do_install_append () {
+	install -d ${D}${sysconfdir}/${PN}
+	install -m 644 ${S}/config.conf.example ${D}${sysconfdir}/${PN}/config.conf
+}


recipes-support/rauc-hawkbit-updater/rauc-hawkbit-updater_git.bb
```diff
@@ -0,0 +1,26 @@
+SUMMARY = "The RAUC hawkBit updater operates as an interface between the RAUC D-Bus API and the hawkBit DDI API."
+HOMEPAGE = "https://github.com/rauc/rauc-hawkbit-updater"
+
+LICENSE = "LGPLv2.1"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=1a6d268fd218675ffea8be556788b780"
+
+SRC_URI = "git://github.com/rauc/rauc-hawkbit-updater.git;protocol=https"
+SRCREV = "84da0e26e5adff7000b60fe22e821570108286a4"
```
ejoerns: Please also set `PV` to a proper value, otherwise it will always be 'git'.
As there is not release yet, maybe an option would be to use:
```
PV = "0.0+git${SRCPV}"
```

recipes-support/rauc-hawkbit-updater/rauc-hawkbit-updater_git.bb
```diff
@@ -0,0 +1,26 @@
...
+S = "${WORKDIR}/git"
+
+inherit cmake pkgconfig systemd
+
+SYSTEMD_SERVICE_${PN} = "rauc-hawkbit-updater.service"
+
+EXTRA_OECMAKE_append = " -DWITH_SYSTEMD=ON"
+
+DEPENDS = "systemd curl glib-2.0-native json-glib"
```
ejoerns: I'd prefer not depending hard on systemd here, especially as rauc-hawkbit-updater explicitly provdes a CMake switch for this.

Thus we can simply make use of the CMake option `WITH_SYSTEMD` and introduce a PACKAGECONFIG for this
The rest we can make depending on `DISTRO_FEATURE` `systemd`:

```suggestion
SYSTEMD_SERVICE_${PN} = "${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'rauc-hawkbit-updater.service', '', d)}"

PACKAGECONFIG ??= "\
    ${@bb.utils.filter('DISTRO_FEATURES', 'systemd', d)} \
    "
PACKAGECONFIG[systemd] = "-DWITH_SYSTEMD=ON,,systemd"

DEPENDS = "curl glib-2.0-native json-glib"
```

recipes-support/rauc-hawkbit-updater/rauc-hawkbit-updater_git.bb
```diff
@@ -0,0 +1,26 @@
...
+
+do_install_append() {
+    install -d ${D}${sysconfdir}/${PN}
+    install -m 644 ${S}/config.conf.example ${D}${sysconfdir}/${PN}/config.conf
+}
+
+FILES_${PN} += " \
```
ejoerns: You can drop this as the `systemd.bbclass` already takes care of it.

recipes-support/rauc-hawkbit-updater/rauc-hawkbit-updater_git.bb
mschwan-phytec: Good point! I will change this.

mschwan-phytec:
Unfortunately, rauc-hawkbit-updater exits with a segmentation fault. Not sure if something is missing in the recipe, or if it is a fault of rauc-hawkbit-updater.

ejoerns:
@mschwan-phytec Is that probably https://github.com/rauc/rauc-hawkbit-updater/issues/21 you encounter? Or in a different situation?

mschwan-phytec:
@ejoerns We tracked this down to the following issue: https://github.com/rauc/rauc-hawkbit-updater/issues/23 So it is not caused by this pull-request.

ejoerns:
@mschwan-phytec I'd say, with updating to the latest master of https://github.com/rauc/rauc-hawkbit-updater, we can merge this first of all and see later if further work on the recipe is required.

In master now both the segfault and the example config file are fixed (https://github.com/rauc/rauc-hawkbit-updater/pull/27, https://github.com/rauc/rauc-hawkbit-updater/pull/25)

mschwan-phytec:
Ok. I updated the SRCREV to the current master.

PR #123, opened 2020-05-20
rauc: fix RRECOMMENDS for rauc-native -> squashfs-tools-native

danc86:
Currently rauc-native RRECOMMENDS squashfs-tools, but it should really
be squashfs-tools-native instead. Normally this distinction is handled
by some magic in native.bbclass, which munges all package relationships
to hack -native onto the end of the names:

https://github.com/openembedded/openembedded-core/blob/master/meta/classes/native.bbclass#L141

but that code only runs if you use BBCLASSEXTEND = "native". If the
recipe explicitly inherits native, as rauc-native does, the renaming
does not happen. Thus the recipe needs to spell out its -native
dependency explicitly.

### DIFF
--- a/recipes-core/rauc/rauc-native.inc
+++ b/recipes-core/rauc/rauc-native.inc
@@ -1,6 +1,7 @@
 inherit native deploy
 
 DEPENDS = "openssl-native glib-2.0-native"
+RRECOMMENDS_${PN} = "squashfs-tools-native"
 
 do_deploy[sstate-outputdirs] = "${DEPLOY_DIR_TOOLS}"
 
--- a/recipes-core/rauc/rauc-target.inc
+++ b/recipes-core/rauc/rauc-target.inc
@@ -1,6 +1,8 @@
 RAUC_KEYRING_FILE ??= "ca.cert.pem"
 RAUC_KEYRING_URI ??= "file://${RAUC_KEYRING_FILE}"
 
+RRECOMMENDS_${PN} = "squashfs-tools"
+
 SRC_URI_append = " \
   file://system.conf \
   ${RAUC_KEYRING_URI} \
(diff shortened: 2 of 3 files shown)


danc86:
Just in case you're wondering why this matters...

Starting from Yocto Zeus, squashfs-tools does not build on musl targets because of some issues with fnmatch: https://git.yoctoproject.org/cgit/cgit.cgi/poky/commit/?id=c6dbd6c52857eb1e59246f337a2d4ab0f3990453

We have a recipe that we build for a musl target which uses rauc-native at build time (it doesn't use rauc at runtime, thankfully). But in Zeus, the recipe now fails to build because of this incorrect RRECOMMENDS from rauc-native to squashfs-tools:

```
ERROR: Nothing RPROVIDES 'squashfs-tools' (but /home/danc/work/platypus-mfg-umbrella/upstream/poky/../meta-rauc/recipes-core/rauc/rauc-native_1.2.bb RDEPENDS on or otherwise requires it)
squashfs-tools was skipped: incompatible with host x86_64-opengear-linux-musl (not in COMPATIBLE_HOST)
NOTE: Runtime target 'squashfs-tools' is unbuildable, removing...
Missing or unbuildable dependency chain was: ['squashfs-tools']
ERROR: Required build target 'coreboot' has no buildable providers.
Missing or unbuildable dependency chain was: ['coreboot', 'rauc-native', 'squashfs-tools']
```

When the dependency is fixed so that it recommends squashfs-tools-native, everything is fine because that's not using musl.

danc86:
I amended the commit to add my signoff.

I'm not sure why the CircleCI build is failing, it requires me to log in and authorize third-party access to my Github account just to see the logs and I don't want to do that.

danc86:
I've amended the commit. Is this what you had in mind?

I left the line in rauc-target.inc:

```
RRECOMMENDS_${PN}_append = " ${PN}-mark-good"
```

untouched, because that seems to fit logically with the chunk of appends at the bottom of the recipe file for handling the rauc-mark-good service. It seemed more logical to just have a straight assignment for `RRECOMMENDS_${PN}` at the top of each file so they mirror each other consistently.

PR #127, opened 2020-06-08
Fix broken master build caused by removed DEPLOY_DIR_TOOLS

ejoerns:
This fixes building meta-rauc master against poky master.

They decided to remove variable `DEPLOY_DIR_TOOLS` and apparently we need to follow this decision to fix broken rauc-native build.

Reference Mail: https://lists.openembedded.org/g/openembedded-core/message/139342

Note this will change rauc host tool installation from tmp/deploy/tools to tmp/deploy/images.

### DIFF
--- a/recipes-core/rauc/rauc-native_1.3.bb
+++ b/recipes-core/rauc/rauc-native_1.3.bb
@@ -5,9 +5,9 @@ inherit native deploy
 do_deploy[sstate-outputdirs] = "${DEPLOY_DIR_TOOLS}"
 
 do_deploy() {
-    install -d ${DEPLOY_DIR_TOOLS}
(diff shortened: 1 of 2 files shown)


ejoerns:
Upstream discussion seems to turn into a direction that they probably revert the patch that broke meta-rauc, thus we will wait a few days for them to decide.

ejoerns:
The patch causing the failure was now finally reverted [1]. Thus we can keep DEPLOY_DIR_TOOLS. I'll make a separate PR to apply the other fix. Needs a rebase/rework onto master anyway.

[1] https://git.yoctoproject.org/cgit/cgit.cgi/poky/commit/?id=d84136cb828847de986a8c627e4782a17b999093

PR #129, opened 2020-06-24
classes: bundle: Add option to build casync bundle

mschwan-phytec:
Add an option to build a casync bundle alongside the regular bundle.
This requires setting an appropriate keyring. Activate the generation of
casync bundles by setting RAUC_CASYNC_BUNDLE to "1".

Signed-off-by: Martin Schwan <m.schwan@phytec.de>

### DIFF
--- a/classes/bundle.bbclass
+++ b/classes/bundle.bbclass
@@ -56,6 +56,10 @@
 #
 #   RAUC_KEY_FILE ?= "development-1.key.pem"
 #   RAUC_CERT_FILE ?= "development-1.cert.pem"
+#
+# For bundle signature verification a keyring file must be provided
+#
+#   RAUC_KEYRING_FILE ?= "ca.cert.pem"
 
 LICENSE = "MIT"
 
@@ -137,6 +141,8 @@ RAUC_KEY_FILE ??= ""
 RAUC_KEY_FILE[doc] = "Specifies the path to the RAUC key file used for signing. Use COREBASE to reference files located in any shared BSP folder."
 RAUC_CERT_FILE ??= ""
 RAUC_CERT_FILE[doc] = "Specifies the path to the RAUC cert file used for signing. Use COREBASE to reference files located in any shared BSP folder."
+RAUC_KEYRING_FILE ??= ""
+RAUC_KEYRING_FILE[doc] = "Specifies the path to the RAUC keyring file used for bundle signature verification. Use COREBASE to reference files located in any shared BSP folder."
 BUNDLE_ARGS ??= ""
 BUNDLE_ARGS[doc] = "Specifies any extra arguments to pass to the rauc bundle command."
 
--- a/classes/bundle.bbclass
+++ b/classes/bundle.bbclass
@@ -52,6 +52,9 @@
 # Extra arguments may be passed to the bundle command with BUNDLE_ARGS eg:
 #   BUNDLE_ARGS += ' --mksquashfs-args="-comp zstd -Xcompression-level 22" '
 #
+# Likewise, extra arguments can be passed to the convert command with
+# CONVERTS_ARGS.
+#
 # Additionally you need to provide a certificate and a key file
 #
 #   RAUC_KEY_FILE ?= "development-1.key.pem"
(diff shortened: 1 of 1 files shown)


classes/bundle.bbclass
```diff
@@ -137,11 +147,14 @@ RAUC_KEY_FILE ??= ""
 RAUC_KEY_FILE[doc] = "Specifies the path to the RAUC key file used for signing. Use COREBASE to reference files located in any shared BSP folder."
 RAUC_CERT_FILE ??= ""
 RAUC_CERT_FILE[doc] = "Specifies the path to the RAUC cert file used for signing. Use COREBASE to reference files located in any shared BSP folder."
+RAUC_KEYRING_FILE ??= ""
+RAUC_KEYRING_FILE[doc] = "Specifies the path to the RAUC keyring file used for signing. Use COREBASE to reference files located in any shared BSP folder."
```
ejoerns, line 151: I'd split this change out as a separate (and extended) commit.

Adding a keyring is also possible for normal signing and will then be used for post-verification. If we already add this, then it should also optionally land in the call of a normal  `rauc sign`.

I'd also slightly adapt the descritpion, maybe

```suggestion
RAUC_KEYRING_FILE[doc] = "Specifies the path to the RAUC keyring file used for bundle signature verification. Use COREBASE to reference files located in any shared BSP folder."
```

mschwan-phytec: Separating the changes seem like a good idea to me.

I realized that there is a `BUNDLE_ARGS` variable that would allow passing the keyring argument to `rauc bundle`. This is a little inconsistent though. If we start adding arguments but allow for specifying custom values with `BUNDLE_ARGS`, things get messy. The same applies for the `rauc convert` command. I guess, at first sight, I should add `BUNDLE_ARGS` there, too? Maybe it would work, if we add the arguments if their respective values are assigned, in the sense like this:
```sh
if ! [ -z "${RAUC_KEYRING_FILE}" ]; then
    BUNDLE_KEYRING_FILE=--keyring="${RAUC_KEYRING_FILE}"
fi

${STAGING_DIR_NATIVE}${bindir}/rauc bundle \
    --debug \
    --cert="${RAUC_CERT_FILE}" \
    --key="${RAUC_KEY_FILE}" \
    ${BUNDLE_KEYRING_FILE} \
    ${BUNDLE_ARGS} \
    ${BUNDLE_DIR} \
    ${B}/bundle.raucb
```
But still, this fights with `BUNDLE_ARGS`. What are your thoughts on this?

ejoerns: I just had a look at the implementation and realized that we introduced `--keyring` and `--signing-keyring` to distinguish between the keyring used only for verification and the one required for conversion / extraction some time ago. Despite `--keyring` actually still works for all cases (for backwarts compatibiliy), we must not (or should not) reflect this in the bundle class. Thus just having an explicit `--keyring=${RAUC_KEYRING_FILE}` for convert should be sufficient.

More interesting the question about `BUNDLE_ARGS`: probably the most important use case is the option to pass `--casync-args` to the convert call. For this, reflecting `BUNDLE_ARGS` in the call would be quite useful. However, I am not fully sure if there are use cases where on might pass some options to the `rauc bundle` call that should not appear in the `rauc convert` call and vice versa. Maybe the safest way to go is to add `CONVERT_ARGS` variable specific for arguments passed to `rauc convert`.

classes/bundle.bbclass
```diff
@@ -301,6 +314,31 @@ do_bundle() {
 		${BUNDLE_ARGS} \
 		${BUNDLE_DIR} \
 		${B}/bundle.raucb
+
+	if [ ${RAUC_CASYNC_BUNDLE} -eq 1 ]; then
+		if [ -z "${RAUC_KEYRING_FILE}" ]; then
+			bbfatal "'RAUC_KEYRING_FILE' not set. Please set a valid keyring file location."
+		fi
+
+		if [ -e ${B}/casync-bundle.raucb ]; then
```
ejoerns, line 323: Maybe this is the point where we should replace the manual cleaning by a 'cleaner'

    do_bundle[cleandirs] = "${B}"

I don't see that there's anything to keep in `${B}` across different runs.

mschwan-phytec:
Rebased and split into multiple commits.

PR #131, opened 2020-07-02
rauc-hawkbit-updater: Add user rauc-hawkbit

mschwan-phytec:
Add a system user "rauc-hawkbit" that is used by the systemd service of rauc-hawkbit-updater.

This fixes https://github.com/rauc/rauc-hawkbit-updater/issues/44.

### DIFF
--- a/recipes-support/rauc-hawkbit-updater/rauc-hawkbit-updater_git.bb
+++ b/recipes-support/rauc-hawkbit-updater/rauc-hawkbit-updater_git.bb
@@ -9,7 +9,7 @@ SRCREV = "d909982e9e4b84cb76b98bf85f25a0a88472301a"
 S = "${WORKDIR}/git"
 PV = "0.0+git${SRCPV}"
 
-inherit cmake pkgconfig systemd
+inherit cmake pkgconfig systemd useradd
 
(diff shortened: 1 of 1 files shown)


recipes-support/rauc-hawkbit-updater/rauc-hawkbit-updater_git.bb
```diff
@@ -18,6 +18,9 @@ PACKAGECONFIG ??= " \
 "
 PACKAGECONFIG[systemd] = "-DWITH_SYSTEMD=ON,,systemd"
 
+USERADD_PACKAGES = "${PN}"
+USERADD_PARAM_${PN} = "--system --home-dir / --no-create-home --shell /bin/nologin rauc-hawkbit"
```
mschwan-phytec, line 22: Is using `/bin/nologin` the correct way to do it? systemd uses this, while others use `/bin/false`. `/bin/nologin` is, at least on my system, not installed.

mschwan-phytec: Changed to `/bin/false` so that trying to switch to this user simply returns exit code 1.

PR #136, opened 2020-07-21
[backport] rauc-hawkbit-updater: Add user rauc-hawkbit

mschwan-phytec:
Add a system user "rauc-hawkbit" that is used by the systemd service of rauc-hawkbit-updater.

---
It would be nice, if we can backport this commit to dunfell. At the moment we cannot directly apply this change to our own layer, since the recipe for rauc-hawkbit-updater exists since dunfell, but this change only since current master.


ejoerns:
I've updated dunfell branch to the current master to include all latest fixes as well as the RAUC 1.4 version bump.
Thus this can be closed as it is already done.

PR #139, opened 2020-08-17
busybox: Add tar autodetect to config

### LINKED ISSUE #128: busybox_%.bbappend doesn't set all configs recomended from docs
In the [rauc docs](https://rauc.readthedocs.io/en/latest/integration.html#required-target-tools) it suggests that the following config options be enabled if using Busybox Tar:
```sh
CONFIG_FEATURE_TAR_AUTODETECT=y
CONFIG_FEATURE_TAR_LONG_OPTIONS=y
```
However, [currently](https://github.com/rauc/meta-rauc/blob/master/recipes-core/busybox/files/rauc.cfg) only `CONFIG_FEATURE_TAR_LONG_OPTIONS` is enabled in the `busybox_%.bbappend` in this layer. Should this also be enabled? I'm happy to open a PR, but wanted to check before putting in the effort.

mcampbellizo:
The RAUC docs recommend setting `CONFIG_FEATURE_TAR_AUTODETECT=y` in
the BusyBox config. This is currently set in upstream Poky in the
default BusyBox config, however it is better to enforce this in
meta-rauc in case that changes upstream.

Fixes #128


ejoerns (approved):
To have this explicit and independent of possible changes in oe-core's default busybox config I agree this is useful.

PR #141, opened 2020-08-26
classes: bundle: Expand bb macros when overriding file name

### LINKED ISSUE #133: No variable expansion in RAUC_SLOT_*[rename]
I have a very non-standard yocto deploy setup and have to use the `RAUC_SLOT_*[rename]` field to bundle the correct images. The value in there is currently not expanded. I would like to set a dynamic value to allow for easy checking if I bundled the correct image.

Problematic LOCs are here:
https://github.com/rauc/meta-rauc/blob/01ee4fa72efa6d97f6f8f25828b8e457b41cdf5b/classes/bundle.bbclass#L238-L239

I have implemented a local fix by replacing 
`imgname = slotflags.get('rename')`
with 
`imgname = d.getVarFlag('RAUC_SLOT_%s' % slot, 'rename')`.

If the current behavior is not intentional I would prepare a quick PR to fix this issue.

Urist-McGit:
When renaming files in a bundle it can be useful to allow for expansion of bitbake macros in the new file name.

Fixes #133

### DIFF
--- a/classes/bundle.bbclass
+++ b/classes/bundle.bbclass
@@ -236,7 +236,7 @@ def write_manifest(d):
             raise bb.build.FuncFailed('Unknown image type: %s' % imgtype)
 
         if slotflags and 'rename' in slotflags:
-            imgname = slotflags.get('rename')
+            imgname = d.getVarFlag('RAUC_SLOT_%s' % slot, 'rename')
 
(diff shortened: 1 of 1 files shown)


ejoerns (approved):
Looks good!

PR #142, opened 2020-09-16
Convert REAMDE to rst and update some chapters

ejoerns:


ejoerns:
You should be able to insect this here: https://github.com/ejoerns/meta-rauc/tree/topic/readme-rst

PR #146, opened 2020-10-02
rauc: rauc-service has a runtime dependency on dbus

ejoerns:
D-Bus is required to communicate with the RAUC service.
Especially on non-systemd distros like poky it could be the case that
there is no D-Bus at all without this explicit runtime dependency.


Emantor (approved):

ejoerns:
backported to dunfell

PR #147, opened 2020-10-02
rauc: split package into 'rauc' for binary and 'rauc-conf' for configuration

ejoerns:
rauc: split package into 'rauc' for binary and 'rauc-conf' for configuration

This allows to build 'rauc' with default TUNE_PKGARCH while only 'rauc-conf' needs to be build for MACHINE_ARCH as it contains MACHINE-specific configuration.

Another point is that the RAUC configuration may not only by MACHINE-specific but may also differ between images (for the same MACHINE).

The rauc target package now `RRECOMMENDS` rauc-conf to pull this in by default.

For providing your own configuration, either bbappend the existing rauc-conf recipe, write a new one that provides 'virtual-rauc-conf' and set `PREFERRED_PROVIDER_virtual-rauc-conf` or just write a custom one and add it to your image's `IMAGE_INSTALL` list (or any packagegroup).

Either way, this requires migration in your rauc recipe(s)!

### DIFF
--- a/recipes-core/rauc/rauc-conf.bb
+++ b/recipes-core/rauc/rauc-conf.bb
@@ -7,6 +7,9 @@ RAUC_KEYRING_URI ??= "file://${RAUC_KEYRING_FILE}"
 
 RPROVIDES:${PN} += "virtual-rauc-conf"
 
+INHIBIT_DEFAULT_DEPS = "1"
+do_compile[noexec] = "1"
+
 SRC_URI = " \
   file://system.conf \
   ${RAUC_KEYRING_URI} \
--- a/recipes-core/rauc/rauc-conf.bb
+++ b/recipes-core/rauc/rauc-conf.bb
@@ -1,0 +1,30 @@
+SUMMARY = "RAUC system configuration & verification keyring"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COREBASE}/meta/COPYING.MIT;md5=3da9cfbcb788c80a0384361b4de20420"
+
+RAUC_KEYRING_FILE ??= "ca.cert.pem"
+RAUC_KEYRING_URI ??= "file://${RAUC_KEYRING_FILE}"
+
+RPROVIDES:${PN} += "virtual-rauc-conf"
+
+SRC_URI = " \
+  file://system.conf \
+  ${RAUC_KEYRING_URI} \
+  "
+
+do_install () {
+        # Create rauc config dir
+        # Warn if system configuration was not overwritten
+        if ! grep -q "^[^#]" ${WORKDIR}/system.conf; then
+                bbwarn "Please overwrite example system.conf with a project specific one!"
+        fi
+        install -d ${D}${sysconfdir}/rauc
+        install -m 0644 ${WORKDIR}/system.conf ${D}${sysconfdir}/rauc/system.conf
+
+        # Warn if CA file was not overwritten
+        if ! grep -q "^[^#]" ${WORKDIR}/${RAUC_KEYRING_FILE}; then
+                bbwarn "Please overwrite example ca.cert.pem with a project specific one, or set the RAUC_KEYRING_FILE variable with your file!"
+        fi
+        install -d ${D}${sysconfdir}/rauc/
+        install -m 0644 ${WORKDIR}/${RAUC_KEYRING_FILE} ${D}${sysconfdir}/rauc/
+}
--- a/recipes-core/rauc/rauc-target.inc
+++ b/recipes-core/rauc/rauc-target.inc
@@ -1,11 +1,10 @@
 RAUC_KEYRING_FILE ??= "ca.cert.pem"
 RAUC_KEYRING_URI ??= "file://${RAUC_KEYRING_FILE}"
 
-RRECOMMENDS:${PN} = "squashfs-tools"
+RRECOMMENDS:${PN} += "virtual-rauc-conf"
+RRECOMMENDS:${PN} += "squashfs-tools"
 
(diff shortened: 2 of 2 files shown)


ejoerns:
rebased onto master

recipes-core/rauc/rauc-conf.bb
```diff
@@ -0,0 +1,31 @@
...
+  "
+
+do_install () {
+        # Create rauc config dir
+        # Warn if system configuration was not overwritten
+        if ! grep -q "^[^#]" ${WORKDIR}/system.conf; then
+                bbwarn "Please overwrite example system.conf with a project specific one!"
+        fi
+        install -d ${D}${sysconfdir}/rauc
+        install -m 0644 ${WORKDIR}/system.conf ${D}${sysconfdir}/rauc/system.conf
```
OnkelUlla, line 23: Perhaps simplify in accordance to the installation of the keyring file:
```suggestion
        install -m 0644 ${WORKDIR}/system.conf ${D}${sysconfdir}/rauc/
```

recipes-core/rauc/rauc-conf.bb
```diff
@@ -0,0 +1,31 @@
+# Copyright (C) 2020 Enrico Jorns <ejo@pengutronix.de>
+# Released under the MIT license (see COPYING.MIT for the terms)
+
+SUMMARY = "RAUC system configuration"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COREBASE}/meta/COPYING.MIT;md5=3da9cfbcb788c80a0384361b4de20420"
+
```
OnkelUlla, line 7: What about
```suggestion

inherit allarch
```
here?

OnkelUlla (commented):
Just two suggestions.

recipes-core/rauc/rauc-conf.bb
```diff
@@ -0,0 +1,31 @@
+# Copyright (C) 2020 Enrico Jorns <ejo@pengutronix.de>
```
jluebbe, line 1:
```suggestion
# Copyright (C) 2021 Enrico Jorns <ejo@pengutronix.de>
```

jluebbe (changes requested):
With the `RDEPENDS += "rauc-conf"` it should be obvious enough to people updating from previous versions that the system.conf would need to move to `rauc-conf`. Perhaps add a hint regarding that to the commit message nevertheless?

recipes-core/rauc/rauc-conf.bb
jluebbe: No, the config needs to be machine-specific.

ejoerns:
Indeed.

And I currently work myself on a project where this should be quite useful, too.
If I find the time, I will test this again in and make it ready for being merged.

philippderdiedas:
well...
how about know? (he asked nicely because does not want to fork it himself)

recipes-core/rauc/rauc-conf.bb
ejoerns: I've simply removed this line.

recipes-core/rauc/rauc-conf.bb
ejoerns: I kept it the same at first, making as few changes as necessary compared to the original recipe.

ejoerns:
Rebased onto the current `master` and applied some minor changes.

I was unsure about how strongly I wanted to couple `rauc` and `rauc-conf` and decided to add a `RRECOMMENDS` to a virtual package indirection since I assume that people not only need to want/need to append the `rauc-conf` recipe but likely provide their own recipes instead. E.g. for differentiating variants on the image level, this is required anyway.

Suggestions or objections welcome.

jluebbe:
> I was unsure about how strongly I wanted to couple `rauc` and `rauc-conf` and decided to add a `RRECOMMENDS` to a virtual package indirection since I assume that people not only need to want/need to append the `rauc-conf` recipe but likely provide their own recipes instead. E.g. for differentiating variants on the image level, this is required anyway.

Anyone using [`NO_RECOMMENDATIONS`](https://docs.yoctoproject.org/singleindex.html#term-NO_RECOMMENDATIONS) in their image will have no config in their image. If you want to use per-image configs, you can easily drop the RDEPENDS via a bbappend. A comment for that case should be enough.

ejoerns:
> > I was unsure about how strongly I wanted to couple `rauc` and `rauc-conf` and decided to add a `RRECOMMENDS` to a virtual package indirection since I assume that people not only need to want/need to append the `rauc-conf` recipe but likely provide their own recipes instead. E.g. for differentiating variants on the image level, this is required anyway.
> 
> With this change, anyone using [`NO_RECOMMENDATIONS`](https://docs.yoctoproject.org/singleindex.html#term-NO_RECOMMENDATIONS) in their image will have no config in their image. If you want to use per-image configs, you can easily drop the RDEPENDS via a bbappend. A comment for that case should be enough.

I wanted to avoid having to bbappend the recipe file for this case and assumed that having no config at all might be noticed.

Note that e.g. systemd pulls in systemd-conf via `RRECOMMENDS` only, too.

But maybe using `RDEPENDS` is safer for the transition in the first round, anyway.

jluebbe:
> I wanted to avoid having to bbappend the recipe file for this case and assumed that having no config at all might be noticed.
> 
> Note that e.g. systemd pulls in systemd-conf via `RRECOMMENDS` only, too.

OK. Then it's consistent, and it retract my criticism.
 
> But maybe using `RDEPENDS` is safer for the transition in the first round, anyway.

jluebbe (approved):

PR #148, opened 2020-10-02
classes: bundle: Avoid using img_fstype from previous iteration

bantu:


ejoerns (approved):
:+1: Good catch!

ejoerns:
Apparently, this commit was untested :/
I've pushed a  [fix](https://github.com/rauc/meta-rauc/commit/04ead4e4968a9fd97511d1bd105c372fc3b2a8c5) to master to add the missing colon after the `else` statement so that bundle class parsing gets fixed.

bantu:
Sorry about that.

PR #153, opened 2020-10-19
bundle.bbclass: fixes for collecting artifacts

ejoerns:
This fixes some long-existing issues with artifact fetching and error handling of bundle.bbclass.

One of them was that when modifying artifact names, an artifact from a previous build could still be in the bundle creation folder and thus in the bundle unintentionally.


Emantor (approved):

ejoerns:
backported to dunfell

PR #155, opened 2020-10-22
bundle: Add slot parameter for bootloader offset

ctenruh-phytec:
Option to set an offset in the slot parameter for the bootloader image. New bootloader image
with the set offset is created via "dd" command.
resolves rauc/rauc#596

### DIFF
--- a/classes/bundle.bbclass
+++ b/classes/bundle.bbclass
@@ -39,6 +39,10 @@
 #   RAUC_SLOT_rootfs ?= "core-image-minimal"
 #   RAUC_SLOT_rootfs[rename] ?= "rootfs.ext4"
 #
+# To prepend an offset to a bootloader image, set the following parameter in bytes.
+# Optionally you can use units allowed by 'dd' e.g. 'K','kB','MB'.
+#   RAUC_SLOT_bootloader[offset] ?= "0"
+#
 # To add additional artifacts to the bundle you can use RAUC_BUNDLE_EXTRA_FILES
 # and RAUC_BUNDLE_EXTRA_DEPENDS.
 # For files from the WORKDIR (fetched using SRC_URI) you can write:
@@ -164,6 +168,7 @@ DEPENDS += "${@bb.utils.contains('RAUC_CASYNC_BUNDLE', '1', 'virtual/fakeroot-na
 
 def write_manifest(d):
     import shutil
+    import subprocess
 
     machine = d.getVar('MACHINE')
     bundle_path = d.expand("${BUNDLE_DIR}")
@@ -240,6 +245,10 @@ def write_manifest(d):
 
         if slotflags and 'rename' in slotflags:
             imgname = d.getVarFlag('RAUC_SLOT_%s' % slot, 'rename')
+        if slotflags and 'offset' in slotflags:
+            imgoffset = slotflags.get('offset')
+            if slotflags.get('offset') == '':
+                imgoffset = '0'
 
         manifest.write("filename=%s\n" % imgname)
         if slotflags and 'hooks' in slotflags:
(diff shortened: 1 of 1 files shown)


classes/bundle.bbclass
```diff
@@ -240,6 +244,11 @@ def write_manifest(d):
 
         if slotflags and 'rename' in slotflags:
             imgname = d.getVarFlag('RAUC_SLOT_%s' % slot, 'rename')
+        if slotflags and 'offset' in slotflags:
+            imgoffset = slotflags.get('offset')
+            if slotflags.get('offset') == "":
+                imgoffset = "0"
+
```
mschwan-phytec, line 251: Remove this additional empty line.

classes/bundle.bbclass
```diff
@@ -252,7 +261,12 @@ def write_manifest(d):
         bb.note("adding image to bundle dir: '%s'" % imgname)
         searchpath = d.expand("${DEPLOY_DIR_IMAGE}/%s") % imgsource
         if os.path.isfile(searchpath):
-            shutil.copy(searchpath, bundle_imgpath)
+            if imgtype == "boot" and 'offset' in slotflags and imgoffset != "0":
+                subprocess.call(["dd", d.expand("if=${DEPLOY_DIR_IMAGE}/%s" % imgsource),
+                                 d.expand("of=${BUNDLE_DIR}/%s") % imgsource,
+                                 "bs=1k", "seek=%s" % imgoffset])
```
mschwan-phytec, line 267: You may want to use `'` instead of `"` for Python strings but that is just a minor thing to me. :wink:

classes/bundle.bbclass
```diff
@@ -252,7 +261,12 @@ def write_manifest(d):
         bb.note("adding image to bundle dir: '%s'" % imgname)
         searchpath = d.expand("${DEPLOY_DIR_IMAGE}/%s") % imgsource
         if os.path.isfile(searchpath):
-            shutil.copy(searchpath, bundle_imgpath)
+            if imgtype == "boot" and 'offset' in slotflags and imgoffset != "0":
+                subprocess.call(["dd", d.expand("if=${DEPLOY_DIR_IMAGE}/%s" % imgsource),
```
mschwan-phytec, line 265: I think you can use `searchpath` instead of `d.expand("if=${DEPLOY_DIR_IMAGE}/%s" % imgsource)` here...

classes/bundle.bbclass
```diff
@@ -252,7 +261,12 @@ def write_manifest(d):
...
+                                 d.expand("of=${BUNDLE_DIR}/%s") % imgsource,
```
mschwan-phytec, line 266: ... and `bundle_imgpath` for this one.

mschwan-phytec (changes requested):

mschwan-phytec (approved):
LGTM

jluebbe:
Is the 1kB granularity small enough? Why not use bytes?

ejoerns:
> Is the 1kB granularity small enough? Why not use bytes?

This can easily be achieved when adding `oflag=seek_bytes` to the dd call.

ctenruh-phytec:
thanks for the hints.

classes/bundle.bbclass
```diff
@@ -39,6 +39,9 @@
 #   RAUC_SLOT_rootfs ?= "core-image-minimal"
 #   RAUC_SLOT_rootfs[rename] ?= "rootfs.ext4"
 #
+# To use an offset in a bootloader image set the following parameter in bytes
```
ejoerns, line 42: `To use an offset` is a bit unspecific. What about `To prepend an offset`?

```suggestion
# To prepend an offset to a bootloader image, set the following parameter in bytes
```

The parameter could potentially also be appended by a common multiplikator as `dd` alows; e.g. `100K`  oder `1MB`

ejoerns (approved):
@ctenruh-phytec Thanks for your adjustments! Looks fine for me now! :+1:

PR #161, opened 2020-11-21
deploy casync native to tools dir

sprhawk:
1. deploy casync native binary to tmp/deploy/tools/casync
1. bump casync source revision to 4ad9bcb94bc83ff36cfc65515107ea06a88c2dfc

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -68,6 +68,10 @@ If you only intend to build the RAUC host tool, you may simply run::
 
 This will place the rauc binary at ``tmp/deploy/tools/rauc``.
 
+If you need to execute the ``casync`` host tool manually, you can do this by running::
+
+  bitbake casync-native -caddto_recipe_sysroot
+  oe-run-native casync-native casync --help
 
 III. Adding the RAUC Update Service to Your Device
 ==================================================


ejoerns:
@sprhawk Thanks for your contirbution. One formal thing: Could you please add a 'Signed-off-by'-line to your commit?

Having casync native tool in tools-deploy dir sounds appropriate to me.

I would structure the recipe slightly different here. What about letting the native recipe simply including the target recipe? Having a zero-line target recipe does not seem to add any value.

And then please remove the BBCLASSEXTEND for native from the base recipe as this would probably overlap with the dedicated target recipe.

sprhawk:
I just did a test as you say.

( delete 'one-line' `casync_git.bb`, rename `casync.inc` to `casync_git.bb`, require `casync_git.bb` at first line in `casync-native_git.bb` )
> I would structure the recipe slightly different here. What about letting the native recipe simply including the target recipe? Having a zero-line target recipe does not seem to add any value.
This will fail as:
```
Did not find CMake 'cmake'
Found CMake: NO
Run-time dependency libzstd found: NO (tried pkgconfig)
meson.build:125:0: ERROR: Dependency "libzstd" not found, tried pkgconfig
```

> And then please remove the BBCLASSEXTEND for native from the base recipe as this would probably overlap with the dedicated target recipe.
This will fail as: 
```
ERROR: Nothing PROVIDES 'udev-native' (but /home/hongbo/Developer/iotpi.io/backupstack/backupstack-yocto/meta-rauc/recipes-casync/casync/casync-native_git.bb DEPENDS on or otherwise requires it). Close matches:
  db-native
  fuse-native
  unifdef-native
ERROR: Required build target 'casync-native' has no buildable providers.
Missing or unbuildable dependency chain was: ['casync-native', 'udev-native']
```

ejoerns:
As using `DEPLOY_DIR_TOOLS` seems to be quite uncommon in Yocto, probably a general alternative would be to run

```
bitbake casync-native -caddto_recipe_sysroot
oe-run-native casync-native casync --help
```

This would at least lower complexity as packages won't need to be split. Any objections on this alternative?

sprhawk:
> As using `DEPLOY_DIR_TOOLS` seems to be quite uncommon in Yocto, probably a general alternative would be to run
> 
> ```
> bitbake casync-native -caddto_recipe_sysroot
> oe-run-native casync-native casync --help
> ```
> 
> This would at least lower complexity as packages won't need to be split. Any objections on this alternative?

I haven't tried it. If it is right, that will work for me. But it'd be better have it documented in README or somewhere

ejoerns:
> I haven't tried it. If it is right, that will work for me. But it'd be better have it documented in README or somewhere

Would you like to make a doc patch? :smirk:

sprhawk:
> > I haven't tried it. If it is right, that will work for me. But it'd be better have it documented in README or somewhere
> 
> Would you like to make a doc patch? smirk

So I will try your method, and make a patch to README. Shall I make a new PR for README patch?

ejoerns:
Just as you like. You can just update this one, too (and change the title) ;)

README.rst
```diff
@@ -68,6 +68,10 @@ If you only intend to build the RAUC host tool, you may simply run::
 
 This will place the rauc binary at ``tmp/deploy/tools/rauc``.
 
+If you need to build ``casync`` to use on host, you may run::
```
ejoerns, line 71: I would at least insert an 'in' after 'to use'.  Maybe a bit clearer is:
``` suggestion
If you need to execute the ``casync`` host tool manually, you can do this by running::
```
as the command below is not primarily for 'building' casync but for executing it.

ejoerns (changes requested):
Please adapt the commit message to have a proper prefix and a comment describing what it actually does. I would also prefer omitting references to the hosting ecosystem from the commit title.

Proposal:

```
README.rst: add hint on how to use casync host tool

[PR #161]

Signed-off-by ...
```

ejoerns (approved):
Looks fine now, thanks! :+1:

PR #164, opened 2020-12-01
added RDEPENDS with casync to rauc

sprhawk:
But this seems needs to set RAUC_CASYNC_BUNDLE = "1" to build/conf/local.conf to be effect

### DIFF
--- a/recipes-core/rauc/rauc-target.inc
+++ b/recipes-core/rauc/rauc-target.inc
@@ -17,6 +17,8 @@ INITSCRIPT_PACKAGES = "${PN}-mark-good"
 INITSCRIPT_NAME_${PN}-mark-good = "rauc-mark-good"
 INITSCRIPT_PARAMS_${PN}-mark-good = "start 99 5 2 . stop 20 0 1 6 ."
 
+RDEPENDS_${PN} += "${@bb.utils.contains('RAUC_CASYNC_BUNDLE', '1', 'casync', '', d)}"
+
 inherit systemd update-rc.d
 
 do_install () {


ejoerns (commented):
The variable `RAUC_CASYNC_BUNDLE` is intended to be used in the bundle recipe. Thus it will not propagate to the global context by default. ThusI'd like to avoid using it here as it will be misleading.

Maybe it is an option to add a `PACKAGECONFIG` option `casync` instead to pull in the RDEPENDS, despite it is a little misuse as there is no such configuration option in rauc.

ejoerns:
@sprhawk As there was no update on this, I'll close this for now.

The general consideration on how to handle runtime dependencies applies to almost all target tools potentially used by rauc (e.g. 'mkfs.ext4', 'resize2fs', 'tar', etc.). The flexibility RAUC does not allow to easily predetermine them. Thus in the end it is the responsibility of the integrator to add the required tools.

I'm open to suggestions on how to solve this in a generic way. But so far, our common recommendation for this is to add the dependencies in the same rauc recipe bbappend where you add the system.conf because this is the place where you decide how you update should behave.

sprhawk:
> @sprhawk As there was no update on this, I'll close this for now.
> 
> The general consideration on how to handle runtime dependencies applies to almost all target tools potentially used by rauc (e.g. 'mkfs.ext4', 'resize2fs', 'tar', etc.). The flexibility RAUC does not allow to easily predetermine them. Thus in the end it is the responsibility of the integrator to add the required tools.
> 
> I'm open to suggestions on how to solve this in a generic way. But so far, our common recommendation for this is to add the dependencies in the same rauc recipe bbappend where you add the system.conf because this is the place where you decide how you update should behave.

Hi I was very busy doing my project and got no time to reply on it. You are right, I agree it is not a good solution. I  just want it can has somewhat dependencies for convenience.

PR #178, opened 2021-02-15
Do not enforce MIT License in bundle.bbclass

hellow554:
Without the `?=` it is not possible to override the License in a derived bb file.
This change is in line with other `.bbclass` files in the poky layer, e.g. `image.bbclass`, `devicetree.bbclass` and similar


ejoerns:
What would be the motivation for using a non-MIT license?

As far as I know this is fully in line with oe-core: https://github.com/openembedded/openembedded/blob/master/classes/image.bbclass#L4

hellow554:
If I have a bundle file (`bundle.bb`) which does `inherit bundle`, and uses a hook via `SRC_URI = "file://hook.sh"`. 

Neither the `bundle.bb` nor `hook.sh` is MIT Licensed (in fact its CLOSED source, because its IP of my company), I can't do `LICENSE="CLOSED"`.

Also it then enforces me to provide a `LIC_FILES_CHKSUM` for my `hook.sh` which is not available.

ejoerns:
> If I have a bundle file (`bundle.bb`) which does `inherit bundle`, and uses a hook via `SRC_URI = "file://hook.sh"`.
> 
> Neither the `bundle.bb` nor `hook.sh` is MIT Licensed (in fact its CLOSED source, because its IP of my company), I can't do `LICENSE="CLOSED"`.

I do not see any reason why a bundle recipe should be IP, but for the hook script, I would agree that this could be an argument, depending on how much logic you put in there.

> Also it then enforces me to provide a `LIC_FILES_CHKSUM` for my `hook.sh` which is not available.

You could point to the proper license file in oe-core, but that is of course only useful if it is actually MIT.

> Interesting enough I was talking about http://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/classes/image.bbclass#n24 which does exactly what I propse in the bbclass file.

Oh, I picked the wrong repo, yes. Anyway, having "?=" not cause us any disadvantage. Thus ok from my side.

@hellow554 would you add your [signed-off-by](https://rauc.readthedocs.io/en/latest/contributing.html#developer-s-certificate-of-origin) to the commit to enable us to merge it?

hellow554:
Done. Thanks for the understanding (:

ejoerns (approved):

ejoerns:
Cherry-picked to 'dunfell'

PR #180, opened 2021-03-01
Add the possibility to add directories to bundle

einoj:
Example usage:
```
SRC_URI_append := " file://hook \
                    file://service.py \
                    file://scripts \
                  "

RAUC_BUNDLE_EXTRA_FILES += "service.py \
                            scripts \
                           "
```
Where scripts is a directory containing files and subdirectories that
will be included in the rauc bundle.

### DIFF
--- a/classes/bundle.bbclass
+++ b/classes/bundle.bbclass
@@ -313,11 +313,19 @@ python do_configure() {
             bb.note("adding extra file from deploy dir to bundle dir: '%s'" % file)
             shutil.copy(searchpath, destpath)
             continue
+        elif os.path.isdir(searchpath):
+            bb.note("adding extra directory from deploy dir to bundle dir: '%s'" % file)
+            shutil.copytree(searchpath, destpath, dirs_exist_ok=True)
+            continue
 
         searchpath = d.expand("${WORKDIR}/%s") % file
         if os.path.isfile(searchpath):
             bb.note("adding extra file from workdir to bundle dir: '%s'" % file)
             shutil.copy(searchpath, destpath)
+            continue
+        elif os.path.isdir(searchpath):
+            bb.note("adding extra directory from workdir to bundle dir: '%s'" % file)
+            shutil.copytree(searchpath, destpath, dirs_exist_ok=True)
             continue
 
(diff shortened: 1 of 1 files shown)


classes/bundle.bbclass
```diff
@@ -313,12 +313,20 @@ python do_configure() {
             bb.note("adding extra file from deploy dir to bundle dir: '%s'" % file)
             shutil.copy(searchpath, destpath)
             continue
+        elif os.path.isdir(searchpath):
+            bb.note("adding extra directory from deploy dir to bundle dir: '%s'" % file)
+            shutil.copytree(searchpath, destpath, dirs_exist_ok=True)
```
ejoerns, line 318: `dirs_exist_ok` was introduced in python3.8 according to https://docs.python.org/3/library/shutil.html#shutil.copytree but bitbake only requires python3.6 currently: http://docs.yoctoproject.org/brief-yoctoprojectqs/index.html#compatible-linux-distribution. I would like to not increase the minimum required version here.

ejoerns:
While reviewing this, I noticed that there are even more things that sould work but don't, e.g. like explicitly specifying paths with subdirectories, like

```
RAUC_BUNDLE_EXTRA_FILES += "testfile testdir/testfile2 testdir/testfile3"
```

Thus I thought about how to fix this in a more generic way and ended up in realizing that this is pretty close to what bitbake fetcher does. Maybe you want to have a look and test the proposal in #192 if that would meet your use case, too?

einoj:
> Maybe you want to have a look and test the proposal in #192 if that would meet your use case, too?

I tried cherry-picking #192 and it works like a charm. I'm closing this PR and will be using #192 instead

makserge:
For rauc 1.12 it is not working.
Rauc documentation has this comment
https://rauc.readthedocs.io/en/stable/faq.html#why-does-bundle-creation-fail-with-a-not-supported-as-contents-error

Do someone knows how to fix this?

ejoerns:
@makserge What do you use the subdirectories in the bundle for?

You might have noticed the comment in the documentation chapter you referened:

> If someone relies on the old undocumented behavior of including directories and symlinks in the bundle, please contact us.

PR #184, opened 2021-03-22
conf: layer: add 'hardknott' to LAYERSERIES_COMPAT

ejoerns:
Remove 'dunfell' that will not be tested with current state.

### DIFF
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -17,7 +17,7 @@ LAYERRECOMMENDS_rauc = "meta-python"
 # meta-filesystems is needed to build cascync with fuse support (the default)
 LAYERRECOMMENDS_rauc += "meta-filesystems"
 
-LAYERSERIES_COMPAT_rauc = "dunfell gatesgarth"
+LAYERSERIES_COMPAT_rauc = "gatesgarth hardknott"
 
 # Sanity check for meta-rauc layer.
 # Setting SKIP_META_RAUC_FEATURE_CHECK to "1" would skip the bbappend files check.


mbessler:
With 'dunfell' being the first Yocto LTS release, do you consider keeping meta-rauc's dunfell branch updated in a similar fashion?

ejoerns:
@mbessler Yes, that's at least my intention. It should have already gained all important patches from master. Currently, I also plan to bump the RAUC versions for dunfell, despite that is a bit against a 'real' stable concept. But my fear is that products with 'dunfell' will be in the field for A LONG TIME. And thus I would like to provide at least a current update mechanism as long as possible.

mbessler:
@ejoerns Thanks, good. 
Yes, dunfell products will likely be in the field quite some time. (In my case, potentially 15 years from now).
As for bumping RAUC versions, I do like having the newer features available, but, at the same time, the upgrade mechanism is one of the most critical pieces of software on a device. 
I wouldn't mind if we had RAUC 1.5 available as long as the dunfell branch exists, at least selectable via the PREFERRED_VERSION mechanism. That would allow products to freeze the RAUC version for the lifetime of the dunfell branch if needed.

ejoerns:
> @ejoerns Thanks, good.
> Yes, dunfell products will likely be in the field quite some time. (In my case, potentially 15 years from now).

That's exactly what I was referring to with *fear*... 
But, If one actually plans to stay on dunfell for 15! years then it really really does not make any difference if once pics a 2 year supported LTS or a 7 month supported release. And I would just hope it is nothing that's 'connected' in any way. :smirk: 

> As for bumping RAUC versions, I do like having the newer features available, but, at the same time, the upgrade mechanism is one of the most critical pieces of software on a device.
> I wouldn't mind if we had RAUC 1.5 available as long as the dunfell branch exists, at least selectable via the PREFERRED_VERSION mechanism. That would allow products to freeze the RAUC version for the lifetime of the dunfell branch if needed.

Note the original RAUC version of dunfell was v1.3. And the current is 1.5.1. And dunfell is no longer supported than until April 2022!

ejoerns:
Note that CI fails due to missing sstate cache which is missing as the master build fails because it lacks this commit.

PR #186, opened 2021-05-04
Fix RAUC_BUNDLE_EXTRA_DEPENDS

hbr-det:
This fixes using RAUC_BUNDLE_EXTRA_DEPENDS resulting in broken
dependencies like recipe:task:task.


ejoerns (approved):
Indeed, this is broken when using the `= recipe:task` syntax. Good catch!

ejoerns:
Cherry-picked to 'dunfell' in #196

PR #187, opened 2021-05-06
classes: bundle: Allow to change Casync bundle name

vkuzmichev:
Let user to customize Casync bundle name using CASYNC_BUNDLE_*NAME set
of variables similarly to BUNDLE_*NAME used for conventional bundles.
New set has similar dependency scheme except that it uses new variable
CASYNC_BUNDLE_BASENAME instead of BUNDLE_BASENAME.

The CASYNC_BUNDLE_BASENAME defaults to "casync-${BUNDLE_BASENAME}" so
this patch makes no changes in bundle names until user would like to
provide custom ones.

Also fix inconsistence in ${B}/casync-bundle.raucb name in do_bundle
and do_deploy tasks.

Signed-off-by: Vitaly Kuzmichev <vkuzmichev@dev.rtsoft.ru>

### DIFF
--- a/classes/bundle.bbclass
+++ b/classes/bundle.bbclass
@@ -326,14 +326,24 @@ python do_configure() {
 do_configure[cleandirs] = "${BUNDLE_DIR}"
 
 BUNDLE_BASENAME ??= "${PN}"
-BUNDLE_BASENAME[doc] = "Specifies desired output base name of generated bundle."
+BUNDLE_BASENAME[doc] = "Specifies desired output base name of generated RAUC bundle."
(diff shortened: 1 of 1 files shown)


ejoerns (changes requested):
Look fine for me, especially that it also fixes the broken BUNDLE_EXTENSION handling. Just some minor notes before we can merge this :+1:

vkuzmichev:
@ejoerns 

Thank you for your feedback. I have pushed new version with suggested amendments.

ejoerns (approved):
:+1:

PR #189, opened 2021-05-18
Remove busybox patch on master (hardknott)

### LINKED ISSUE #188: Busybox patch does not apply on hardknott
PR #151 introduced a patch for busybox that got integrated with busybox version 1.33.0. Poky ships busybox with version 1.33.0 since hardknott. As a result, on hardknott, meta-rauc does not build because the patch does not apply.

Now, I thought the easiest way to resolve this is to simply remove this patch for master (currently hardknott). But I noticed that the layer is marked as compatible for both gatesgarth and hardknott. On gatesgarth this patch is needed, because busybox has version 1.32.0.

What is the proposed solution here? Is removing compatibility for gatesgarth on the master branch and deleting the patch on master an option? (Because there is a dedicated branch for gatesgarth anyway)

mschwan-phytec:
Remove a busybox patch from master that got upstreamed and remove layer compatibility for gatesgarth on master to avoid conflicts with that patch.

Resolves #188.


jkroon81 (approved):
Looks good to me

Emantor (approved):

PR #191, opened 2021-06-11
Added support for dm-verity into kernel config

rforro:
Current kernel.cfg is missing support for mounting of dm-verity devices, this configuration fixes it.


ejoerns (changes requested):
Thanks for your PR. Yes, this should be enabled here, too of course.

However, I would boild down config fragment lines to what is actually required for verity to work. Having all these 'no set' lines in the fragment will otherwise lead to having features unnecessarily disabled. This is probably not our intention.

Note that unfortunately, this is obviously wrong in the already existing part of the config file, too.

For verity the two lines that count (as they are not automatically selected) are:

    CONFIG_MD
    CONFIG_BLK_DEV_DM=y
    CONFIG_DM_VERITY=y

Feel free to also fix this for the other (existing) settings if you like.

rforro:
No problem, I'll change it. Should I remove those `is not set` in the existing config file as well or better not mix it with this PR?

The second thing what I don't understand is why by using `menuconfig` to activate verity it enables so many modules if only those two are needed for dm-verity?

When I run diff on before and after activation I get something like this:
```
-- og.config   2021-06-30 16:17:34.150684994 +0200
+++ verity.config       2021-06-30 16:21:24.166682012 +0200
@@ -1595,7 +1595,37 @@
 # end of SCSI device support
 
 # CONFIG_ATA is not set
-# CONFIG_MD is not set
+CONFIG_MD=y
+# CONFIG_BLK_DEV_MD is not set
+# CONFIG_BCACHE is not set
+CONFIG_BLK_DEV_DM_BUILTIN=y
+CONFIG_BLK_DEV_DM=y
+# CONFIG_DM_DEBUG is not set
+CONFIG_DM_BUFIO=y
+# CONFIG_DM_DEBUG_BLOCK_MANAGER_LOCKING is not set
+# CONFIG_DM_UNSTRIPED is not set
+# CONFIG_DM_CRYPT is not set
+# CONFIG_DM_SNAPSHOT is not set
+# CONFIG_DM_THIN_PROVISIONING is not set
+# CONFIG_DM_CACHE is not set
+# CONFIG_DM_WRITECACHE is not set
+# CONFIG_DM_ERA is not set
+# CONFIG_DM_CLONE is not set
+# CONFIG_DM_MIRROR is not set
+# CONFIG_DM_RAID is not set
+# CONFIG_DM_ZERO is not set
+# CONFIG_DM_MULTIPATH is not set
+# CONFIG_DM_DELAY is not set
+# CONFIG_DM_DUST is not set
+# CONFIG_DM_INIT is not set
+# CONFIG_DM_UEVENT is not set
+# CONFIG_DM_FLAKEY is not set
+CONFIG_DM_VERITY=y
+# CONFIG_DM_VERITY_VERIFY_ROOTHASH_SIG is not set
+# CONFIG_DM_VERITY_FEC is not set
+# CONFIG_DM_SWITCH is not set
+# CONFIG_DM_LOG_WRITES is not set
+# CONFIG_DM_INTEGRITY is not set
 # CONFIG_TARGET_CORE is not set
 CONFIG_NETDEVICES=y
 CONFIG_MII=m
@@ -4949,6 +4979,7 @@
 CONFIG_NFS_V4_1=y
 CONFIG_NFS_V4_2=y
 CONFIG_PNFS_FILE_LAYOUT=y
+CONFIG_PNFS_BLOCK=y
 CONFIG_PNFS_FLEXFILE_LAYOUT=m
 CONFIG_NFS_V4_1_IMPLEMENTATION_ID_DOMAIN="kernel.org"
 # CONFIG_NFS_V4_1_MIGRATION is not set
```

ejoerns:
@rforro This basically is caused by how kconfig works. We see the result of two things here:

1) Some features require other features to be enabled. For this, a Kconfig entry can itself `select` additional Kconfig entries. The `DM_VERITY` entry for example looks as follows:

```
config DM_VERITY
        tristate "Verity target support"
        depends on BLK_DEV_DM
        select CRYPTO
        select CRYPTO_HASH
        select DM_BUFIO
        help
        [...]
```

2) Features can be grouped into sub menus in kconfig. Menus that are not selected at all do not show their subitems. This is true both for the interactive GUI tools as well as for the resulting config file. In our case we enable for example `CONFIG_MD`. The menu roughly looks as follows:

```
menuconfig MD
        bool "Multiple devices driver support (RAID and LVM)"
        depends on BLOCK
        select SRCU
        help
          Support multiple physical spindles through a single logical device.
          Required for RAID and logical volume management.

if MD

config BLK_DEV_MD
        [...]

config MD_AUTODETECT
        [...]

[...]

endif # MD

```

Thus by enabling CONFIG_MD, also all menu items hidden before now become visible and get their default value. This is what makes the actual diff that long. Also note that most of the added options are not enabled actually.

ejoerns (changes requested):
@rforro The result already looks almost fine now, thanks!

But I would like to avoid PRs that add add thing in one commit that they remove in the following.

What I propose are 2 commit in the following order (and with a proper commit subject prefix):

1. "linux-yocto: ￼removed unnecessary default kernel config options"

    -> This one removes everything from the current master config fragment that will not be set by kconfig default values, anyway.

  This should boil down to keeping only:

    CONFIG_BLK_DEV_LOOP=y
    CONFIG_SQUASHFS=y

2. "linux-yocto: add support for dm-verity into kernel config"

  Which adds only these lines:

    CONFIG_MD=y
    CONFIG_BLK_DEV_DM=y
    CONFIG_DM_VERITY=y



Not that not only the "is not set" lines are defaults, but some of the "=y" lines, too. A value can default either to being unset or to being set.

rforro:
@ejoerns done

I have only one question according to option `CONFIG_SQUASHFS_FILE_CACHE`. How we know, that this will be selected? Is it because it's the first `config ` in `choice` list?

```
choice
	prompt "File decompression options"
	depends on SQUASHFS
	help
	  --omitted--

config SQUASHFS_FILE_CACHE
	bool "Decompress file data into an intermediate buffer"
	help
	  --omitted--

config SQUASHFS_FILE_DIRECT
	bool "Decompress files directly into the page cache"
	help
	  --omitted--

endchoice
```

ejoerns:
@rforro cannot find this explicitly in kconfig documentation but I would assume that a choice defaults to its first element, yes. Maybe if a kernel experts reads this, he can give a hint if that's documented in text anywhere or only in code ;)

Note that the commit for master is now rejected because they have moved the next code name ('honister'). Thus we will need to wait for #195 and rebase on this then.

ejoerns:
As #195 is merged now, feel free to rebase onto master so we can make this ready for being merged

ejoerns:
I've rebased this on latest master

jluebbe:
We should also add `CONFIG_CRYPTO_SHA256=y` to the config.

ejoerns:
Added `CONFIG_CRYPTO_SHA256=y`

jluebbe (approved):

ejoerns (approved):

PR #192, opened 2021-06-25
classes: bundle: rework fetcher logic to handle dirs / subdirs etc.

ejoerns:
This is inspired by what the bitbake fetcher code does and allows to copy directories, files in directories by preserving the full paths,
etc.


einoj:
This solves #180. Would be great to have this merged into dunfell as well!

PR #193, opened 2021-06-25
classes: bundle: add support to find imagesource in subdir

Jarsop:
In some cases the image to include in RAUC bundle
may be inside a $DEPLOY_DIR_IMAGE subdirectory
and imgname variable will contains it in the path.
The result is a failure at the image copy and the
filename manifest field also contains it.
This change wrap the imgname with a Python PurePath
to keep only the image filename instead of full path.

Signed-off-by: Jarsop <jarsop@outlook.com>

### DIFF
--- a/classes/bundle.bbclass
+++ b/classes/bundle.bbclass
@@ -176,6 +176,7 @@ DEPENDS += "${@bb.utils.contains('RAUC_CASYNC_BUNDLE', '1', 'virtual/fakeroot-na
 def write_manifest(d):
     import shutil
     import subprocess
+    from pathlib import PurePath
 
     machine = d.getVar('MACHINE')
     bundle_path = d.expand("${BUNDLE_DIR}")
@@ -262,6 +263,8 @@ def write_manifest(d):
             if slotflags.get('offset') == '':
                 imgoffset = '0'
 
+        # Keep only the image name in case the image is in a $DEPLOY_DIR_IMAGE subdirectory
+        imgname = PurePath(imgname).name
         manifest.write("filename=%s\n" % imgname)
         if slotflags and 'hooks' in slotflags:
             if not have_hookfile:


classes/bundle.bbclass
```diff
@@ -262,15 +263,16 @@ def write_manifest(d):
             if slotflags.get('offset') == '':
                 imgoffset = '0'
 
-        manifest.write("filename=%s\n" % imgname)
+        imgname = PurePath(imgname)
```
ejoerns, line 266: As we only intend to use `imgname.name`, we could replace this by

``` suggestion
        imgname = PurePath(imgname).name
```

and then refer `imagename` only. This would also prevent from changing the type of `imgname`.

Maybe also add a comment on that line as the purpose is possibly not fully intuitive.

Jarsop: Good point, my original purpose coming from older meta-rauc version. So it's done. Thanks for your review

ejoerns (approved):
I've pushed some minor typo fixes. Looks fine for me now :+1:

Jarsop:
Do you want me to backport this purpose on other branches (warrior, zeus, dunfell, gatesgarth) ?

Jarsop:
And maybe you want me to rebase on master to be aligned with latest changes ?

ejoerns:
> @ejoerns Do you want me to backport this purpose on other branches (warrior, zeus, dunfell, gatesgarth) ? Just a picking will be enougth

@Jarsop I marked it as a backport candidate, thx. No need to make a separate PR, as I will do backports from time to time when things proved to be stable on master.

> And maybe you want me to rebase on master to be aligned with latest changes ?

For single-commit PR's we do a 'Rebase and merge' anyway. Thus there should be no need to rebase as long as it is conflict free and we don't expect testing to produce quite different results on master (which should not be the case here I guess).

Jarsop:
Perfect, I don't know how you test integrations, but you can add a test case that builds a bundle from an image that provides from a `$ DEPLOY_DIR_IMAGE` subdirectory now: wink:

See you at the next contribution I hope: slightly_smiliing_face :,

ejoerns:
@Jarsop (Automated) testing is still quite unsophisticated but more than we had before at least. The testing is done via GitHub Actions, defined in https://github.com/rauc/rauc/blob/master/.github/workflows/tests.yml

Jarsop:
@ejoerns Ok but maybe I'm not smart about it, but I don't see anything related to Yocto bundle build in this GitHub workflow

ejoerns:
@Jarsop Oh, I referenced the wrong one (the one from rauc itself). The right one is: https://github.com/rauc/meta-rauc/blob/master/.github/workflows/meta-rauc.yml But bundle build is still missing there obviously (despite we should be able to build at least `core-bundle-minimal`. Most of the feature-specific testing of I do manually, anyway. The tests are meant to assure that we do not break existing functionality. But as I said, this is only the initial setup. Took some time to get it working halfway stable with acceptable build times and sstate pre-population, etc.

PR #195, opened 2021-08-08
convert to new override syntax

MarkusVolk:
Signed-off-by: Markus Volk <f_l_k@t-online.de>

I can't promise I haven't missed anything, but hopefully there shouldn't be any false positives

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -84,8 +84,8 @@ you have to follow at least the following steps:
    ``/etc/rauc/system.conf``. For information on how to write the RAUC
    update file, please refer to the RAUC user documentation [1]_::
 
-     FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
-     SRC_URI_append := " file://system.conf"
+     FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
+     SRC_URI:append := " file://system.conf"
 
 3. Create a bundle recipe for your device by adding a recipe
    that inherits the `bundle` class and adds all desired
@@ -112,7 +112,7 @@ you have to follow at least the following steps:
 Note: If you do not use packagegroup-base, you als need to manually add
 the `rauc` package to your systems image recipe::
 
-     IMAGE_INSTALL_append = " rauc"
+     IMAGE_INSTALL:append = " rauc"
 
 
 IV. Building The RAUC hawkBit Clients
@@ -127,7 +127,7 @@ hawkBit deployment server:
 To use ``rauc-hawkbit`` as a standalone service add to your systems image
 recipe::
 
-    IMAGE_INSTALL_append = " rauc-hawkbit-service"
+    IMAGE_INSTALL:append = " rauc-hawkbit-service"
 
 To use it as a python library in your demo application instead, simply add to
 your recipe::
@@ -136,7 +136,7 @@ your recipe::
 
 To use ``rauc-hawkbit-updater`` in your system add to your image recipe::
 
-    IMAGE_INSTALL_append = " rauc-hawkbit-updater"
+    IMAGE_INSTALL:append = " rauc-hawkbit-updater"
 
 V. Configure Custom Kernel
 ==========================
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -17,7 +17,7 @@ LAYERRECOMMENDS_rauc = "meta-python"
 # meta-filesystems is needed to build cascync with fuse support (the default)
 LAYERRECOMMENDS_rauc += "meta-filesystems"
 
-LAYERSERIES_COMPAT_rauc = "hardknott"
+LAYERSERIES_COMPAT_rauc = "honister"
 
 # Sanity check for meta-rauc layer.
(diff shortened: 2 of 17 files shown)


conf/layer.conf
```diff
@@ -17,7 +17,7 @@ LAYERRECOMMENDS_rauc = "meta-python"
 # meta-filesystems is needed to build cascync with fuse support (the default)
 LAYERRECOMMENDS_rauc += "meta-filesystems"
 
-LAYERSERIES_COMPAT_rauc = "hardknott"
+LAYERSERIES_COMPAT_rauc = "hardknott honister"
```
texierp, line 20: Hi @MarkusVolk 

I think that we need to remove 'kardknott' from `LAYERSERIES_COMPAT` . This release is not compatible the new override syntax.

MarkusVolk: Hi @texierp 

I thouhgt that the new syntax would be backward compatible and only master branch doesn't accept old syntax anymore. But don't nail me down on this. If you like, i will remove hardknott

texierp: In fact, at the moment the new syntax is not fully backward compatible, there are bitbake's patches for older releases, and if they are applied, both "_"  and ":" are recognized as override chars. Only Poky's dunfell contains this patch, see [1]

[1] - https://git.yoctoproject.org/cgit.cgi/poky/log/?h=dunfell

jluebbe:
Did you do the changes manually or with the conversion script?

MarkusVolk:
I used some oneliners with for loop, find and sed, but very conservatively to not generate false positives and only did this for .bb and .bbappend files. After that i manually looked at the result and changed what i was missing. Enough to make it buildable with current master branch but bundle.bbclass still needs to be converted.

Found out about the conversion script just after. I dont have any experience how good this thing works

MarkusVolk:
Just did run convert.py on meta-rauc. It also doesn't do changes to bundle.bbclass but discovered two missing converts

```
diff --git a/recipes-core/packagegroups/packagegroup-base_rauc.inc b/recipes-core/packagegroups/packagegroup-base_rauc.inc
index 30a6976..35d26f5 100644
--- a/recipes-core/packagegroups/packagegroup-base_rauc.inc
+++ b/recipes-core/packagegroups/packagegroup-base_rauc.inc
@@ -1,6 +1,6 @@
 PACKAGES += "packagegroup-base-rauc"
 RDEPENDS:packagegroup-base += "packagegroup-base-rauc"
 
-SUMMARY_packagegroup-base-rauc = "RAUC update framework support"
+SUMMARY:packagegroup-base-rauc = "RAUC update framework support"
 RDEPENDS:packagegroup-base-rauc = "\
     rauc"
diff --git a/recipes-support/rauc-hawkbit-updater/rauc-hawkbit-updater_git.bb b/recipes-support/rauc-hawkbit-updater/rauc-hawkbit-updater_git.bb
index 5037913..3f8d0a4 100644
--- a/recipes-support/rauc-hawkbit-updater/rauc-hawkbit-updater_git.bb
+++ b/recipes-support/rauc-hawkbit-updater/rauc-hawkbit-updater_git.bb
@@ -11,7 +11,7 @@ PV = "0.0+git${SRCPV}"
 
 inherit cmake pkgconfig systemd useradd
 
-SYSTEMD_SERVICE_${PN} = "${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'rauc-hawkbit-updater.service', '', d)}"
+SYSTEMD_SERVICE:${PN} = "${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'rauc-hawkbit-updater.service', '', d)}"
 
 PACKAGECONFIG ??= " \
     ${@bb.utils.filter('DISTRO_FEATURES', 'systemd', d)} \
(END)

```

MarkusVolk:
I guess in bundle.bbclass we need to replace every `RAUC_SLOT_%s` with `RAUC_SLOT:%s`?

jluebbe:
> I guess in bundle.bbclass we need to replace every `RAUC_SLOT_%s` with `RAUC_SLOT:%s`?

See https://github.com/rauc/meta-rauc/blob/master/classes/bundle.bbclass#L133. There it's fixed to `_%s`, but as it's not really a normal override, I wouldn't change that to `:`.

README.rst
```diff
@@ -84,12 +84,12 @@ you have to follow at least the following steps:
...
 
-     FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
-     SRC_URI_append := " file://system.conf"
+     FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
+     SRC_URI:append := " file://system.conf"
 
 3. Create a bundle recipe for your device by adding a recipe
    that inherits the `bundle` class and adds all desired
-   configuration::
+   configuration:
```
ejoerns, line 92: The double-colon should not be removed here as it introduces a code block.

ejoerns (approved):
I've fixed the minor documentation issue. Otherwise, it looks fine for me. Thanks! :+1:

PR #197, opened 2021-08-23
rauc: rauc-mark-good: improve service unit

Bastian-Krause:
Make the rauc-mark-good.service run only if `bootchooser.active` or `rauc.slot` are set in the kernel command line. During
debugging/development, boot from NFS or non-bootchooser boots might be used. In these cases `rauc status mark-good` will fail.

`RemainAfterExit=yes` allows systemd to start dependencies only after `rauc status mark-good` has succeeded.

While at it, use the official RAUC spelling and remove the trailing empty line.

### DIFF
--- a/recipes-core/rauc/files/rauc-mark-good.service
+++ b/recipes-core/rauc/files/rauc-mark-good.service
@@ -1,12 +1,14 @@
 [Unit]
-Description=Rauc Good-marking Service
+Description=RAUC Good-marking Service
+ConditionKernelCommandLine=|bootchooser.active
+ConditionKernelCommandLine=|rauc.slot
 After=boot-complete.target
 Requires=boot-complete.target
 
 [Service]
 Type=oneshot
+RemainAfterExit=yes
 ExecStart=@BINDIR@/rauc status mark-good
 
 [Install]
 WantedBy=multi-user.target
-


OnkelUlla (changes requested):
Good idea to utilize ``ConditionKernelCommandLine=|...`` to start the unit only if the bootloader explicitely booted one or the other slot. But then the directive ``WantedBy=multi-user.target`` should be removed at the same time.

The introduction of ``RemainAfterExit=yes`` makes sense as well because it ensures that this service unit is only started once (see e.g. section "Example 4. Stoppable oneshot service" in the manpage for [systemd.service](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Examples)). Starting dependencies only after success of ``rauc status mark-good`` is already arranged by the choice of ``Type=oneshot`` compared to ``Type=simple``, so the commit message should be adjusted in that respect.

jluebbe:
> Good idea to utilize `ConditionKernelCommandLine=|...` to start the unit only if the bootloader explicitely booted one or the other slot. But then the directive `WantedBy=multi-user.target` should be removed at the same time.

I think `WantedBy` is still needed to activate the unit. Are you sure this works without it?

Bastian-Krause:
> The introduction of `RemainAfterExit=yes` makes sense as well because it ensures that this service unit is only started once (see e.g. section "Example 4. Stoppable oneshot service" in the manpage for [systemd.service](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Examples)). Starting dependencies only after success of `rauc status mark-good` is already arranged by the choice of `Type=oneshot` compared to `Type=simple`, so the commit message should be adjusted in that respect.

Correct, I misread the man page regarding that option. Will adjust the commit message.

OnkelUlla:
> I think `WantedBy` is still needed to activate the unit. Are you sure this works without it?

You are right and my understanding of triggering conditions was not correct, so please forget about my suggestion to remove ``WantedBy=multi-user.target``.

OnkelUlla (approved):
Looks good to me!

ejoerns (approved):

ejoerns:
Cherry-picked to 'dunfell' branch.

PR #201, opened 2021-10-13
Add support for full custom update

vivien:
This PR adds support for full custom updates as well as simplifying the code a bit and enabling variable parsing in flags such as `RAUC_SLOT_rootfs[file] = "${SOME_FILE}"`.


vivien:
Any update on this? :eyes:

classes/bundle.bbclass
```diff
@@ -398,13 +407,15 @@ inherit deploy
...
 	install -m 0644 ${B}/bundle.raucb ${DEPLOYDIR}/${BUNDLE_NAME}${BUNDLE_EXTENSION}
 	ln -sf ${BUNDLE_NAME}${BUNDLE_EXTENSION} ${DEPLOYDIR}/${BUNDLE_LINK_NAME}${BUNDLE_EXTENSION}
 
 	if [ ${RAUC_CASYNC_BUNDLE} -eq 1 ]; then
-		install ${B}/casync-bundle.raucb ${DEPLOYDIR}/${CASYNC_BUNDLE_NAME}${CASYNC_BUNDLE_EXTENSION}
-		cp -r ${B}/casync-bundle.castr ${DEPLOYDIR}/${CASYNC_BUNDLE_NAME}.castr
+		install -m 0644 ${B}/casync-bundle.raucb ${DEPLOYDIR}/${CASYNC_BUNDLE_NAME}${CASYNC_BUNDLE_EXTENSION}
 		ln -sf ${CASYNC_BUNDLE_NAME}${CASYNC_BUNDLE_EXTENSION} ${DEPLOYDIR}/${CASYNC_BUNDLE_LINK_NAME}${CASYNC_BUNDLE_EXTENSION}
+
+		cp --no-preserve=ownership -r ${B}/casync-bundle.castr ${DEPLOYDIR}/${CASYNC_BUNDLE_NAME}.castr
```
ejoerns, line 418: Since this is nothing that is installed on the target, I am not sure if how host contamination plays a role here.

classes/bundle.bbclass
```diff
@@ -300,7 +309,7 @@ python do_configure() {
 
     write_manifest(d)
 
-    hooksflags = d.getVarFlags('RAUC_BUNDLE_HOOKS')
+    hooksflags = d.getVarFlags('RAUC_BUNDLE_HOOKS', ['file', 'hooks'])
```
ejoerns, line 312: Didn't knew this notation is possible, yet. I like this solution since it does not require to add additional `getVarFlag()` calls.

classes/bundle.bbclass
```diff
@@ -111,6 +111,7 @@ RAUC_BUNDLE_DESCRIPTION[doc] = "Specifies the bundle description string. See RAU
 RAUC_BUNDLE_BUILD[doc] = "Specifies the bundle build stamp. See RAUC documentation for more details."
 
 RAUC_BUNDLE_SLOTS[doc] = "Space-separated list of slot classes to include in bundle (manifest)"
+RAUC_BUNDLE_HANDLER[doc] = "Allows to fully replace the default update process with a custom script (via varflags '[filename'] and ['args'])"
```
ejoerns, line 114: This is a really rarely used feature. Thus if we support it in the manifest, it should at least get a name that does not confuse with normal RAUC handlers and that clearly says you're doing something very custom.

Thus maybe `RAUC_BUNDLE_FULL_CUSTOM_HANDLER` or at least `RAUC_BUNDLE_CUSTOM_HANDLER`?

classes/bundle.bbclass
```diff
@@ -148,10 +150,9 @@ python __anonymous() {
             d.appendVarFlag('do_unpack', 'depends', ' ' + image + ':do_deploy')
 
     for image in (d.getVar('RAUC_BUNDLE_EXTRA_DEPENDS') or "").split():
-        imagewithdep = image.split(':')
-        deptask = imagewithdep[1] if len(imagewithdep) > 1 else 'do_deploy'
-        d.appendVarFlag('do_unpack', 'depends', ' %s:%s' % (imagewithdep[0], deptask))
-        bb.note('adding extra dependency %s:%s' % (imagewithdep[0],  deptask))
+        pn, task = ('%s:do_deploy' % image).split(':')[0:2]
```
ejoerns, line 153: this is way shorter but maybe a bit too magic to get at first sight later what we did here.
Could you add a short comment to clarify how 'fallback' to `do_deploy` works, here?

ejoerns (changes requested):
Sorry for letting this unattended for longer. I was mainly a bit skeptical about making support for full custom updates too convenient since I don't really think that this is what users should do in 90% of use cases.
Over this the bug fixes contained here popped of my mental stack. But to be honest, they are unrelated to the title, anyway.

Thus I would propose to create a separate PR for the fixes and keep this PR for the full custom handler.

I would tend to merge full custom handler, too if the adaption I noted was made.

ejoerns:
Since most code of the `bundle.bbclass` is useless for full custom bundles, I find it misleading for others to implement the full custom update here. A distinct class might be a better solution.

Since the other fixes that were part of this PR were resolved by later PRs already, I'd close this now.
@vivien sorry for ignoring the useful fixes that were in here for too long.

PR #206, opened 2021-11-05
recipes: Update SRC_URI branch and protocols

marcel-hamer:
Update the SRC_URIs to include a branch and to include a protocol for
GitHub. This was performed using the convert script from
openembedded-core.

Signed-off-by: Marcel Hamer <marcel@solidxs.se>

### DIFF
--- a/recipes-casync/casync/casync_git.bb
+++ b/recipes-casync/casync/casync_git.bb
@@ -9,7 +9,7 @@ SRCREV = "a8f6c841ccfe59ca8c68aad64df170b64042dce8"
 PV = "2+git${SRCPV}"
 
 SRC_URI = " \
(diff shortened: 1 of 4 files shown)


jluebbe (approved):

Bastian-Krause (approved):

ejoerns (approved):
Although the commit claims to do so, it does not actually change protocols here.
But as it is the intention of the 'script' (namely https://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/scripts/contrib/convert-srcuri.py) to insert both protocol and branch name when not being present yet, I'll accept this.

ejoerns:
@marcel-hamer Thanks for your first contribution!

PR #207, opened 2021-11-05
classes: bundle: support multiconfig dependencies for slots

a3f:
Users may want to populate a slot with an image resulting from a
multiconfig build. This can be especially useful with RAUC variants:

 - Same rootfs slots for both `MACHINE`s
 - different bootloader images, depending on `MACHINE`

For this to work out of the box, `bundle.bbclass` will need some
multiconfig awareness to emit `mcdepends` instead of `depends`
when depending on a multiconfig recipe.

With this addressed, following slots definition is now possible:
```
  RAUC_SLOT_bootloader-othermachine = "mc::othermachine:bootloader"
  RAUC_SLOT_bootloader-othermachine[name] = "bootloader.othermachine"
  RAUC_SLOT_bootloader-othermachine[type] = "boot"
  RAUC_SLOT_bootloader-othermachine[file] = "../othermachine/bootloader.img"
```
No functional change intended for non-multiconfig users.

### DIFF
--- a/classes/bundle.bbclass
+++ b/classes/bundle.bbclass
@@ -123,6 +123,14 @@ RAUC_BUNDLE_FORMAT[doc] = "Specifies the bundle format to be used (plain/verity)
 
 # Create dependency list from images
 python __anonymous() {
+    def append_slot_depends(d, image, deptask=None):
+        dependency = "%s:%s" % (image, deptask) if deptask else image
+
+        if image.startswith('multiconfig:') or image.startswith('mc:'):
+            d.appendVarFlag('do_unpack', 'mcdepends', ' ' + dependency)
+        else:
+            d.appendVarFlag('do_unpack', 'depends', ' ' + dependency)
+
     d.appendVarFlag('do_unpack', 'vardeps', ' RAUC_BUNDLE_HOOKS')
     for slot in (d.getVar('RAUC_BUNDLE_SLOTS') or "").split():
         slotflags = d.getVarFlags('RAUC_SLOT_%s' % slot)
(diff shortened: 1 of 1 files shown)


a3f:
I have the equivalent on zeus and it works well. I left this as a draft until I can test this against a more up to date Yocto release. Comments welcome in the meantime though!

classes/bundle.bbclass
```diff
@@ -123,6 +123,14 @@ RAUC_BUNDLE_FORMAT[doc] = "Specifies the bundle format to be used (plain/verity)
 
 # Create dependency list from images
 python __anonymous() {
+    def append_slot_depends(d, image, deptask=None):
```
ejoerns, line 126: Maybe rename `image` to `depends` here to avoid confusion since it is not necessarily an 'image' and it might also contain the deptask already.

classes/bundle.bbclass
```diff
@@ -123,6 +123,14 @@ RAUC_BUNDLE_FORMAT[doc] = "Specifies the bundle format to be used (plain/verity)
...
+        dependency = "%s:%s" % (image, deptask) if deptask else image
+
+        if image.startswith('multiconfig:') or image.startswith('mc:'):
+            d.appendVarFlag('do_unpack', 'mcdepends', ' ' + dependency)
```
ejoerns, line 130: I guess the `' '` is not required since we have a string already.

ejoerns (changes requested):

a3f:
In the meantime, I have added an intermediary recipe that does basically:
```
OTHERMACHINE_DEPLOYDIR= "../../../../tmp_mc-othermachine/deploy/images/othermachine"
do_deploy[mcdepends] += "mc::othermachine:originalrecipe:do_deploy
do_deploy () {
  cd ${DEPLOY_DIR_IMAGE}
  ln -sf ${OTHERMACHINE_DEPLOYDIR}/artifact-othermachine $ ${DEPLOYDIR}/
}
```

So I am no longer using the changes I originally submitted here. I will try to find some time in a few weeks to incorporate your feedback and retest.

ejoerns:
@a3f  Any update on this or interest in the topic? Or should we close this until someone needs mc deps again?

a3f:
Sorry, didn't find the time to address the feedback. As I don't currently use it, I'll close till there's renewed interest.

PR #209, opened 2021-11-26
update dt-utils to 2021.03.0

hellow554:
Fixes https://bugzilla.yoctoproject.org/show_bug.cgi?id=14186

(also removes the patch, because it was upstreamed)


hellow554:
There seems to be a missing part in the CI, that the missing checksum was not detected...

Should be investigated ;)

ejoerns:
> There seems to be a missing part in the CI, that the missing checksum was not detected...
> 
> Should be investigated ;)

No need to. Building dt-utils was simply not covered by the CI script yes :smirk: 

I pushed an update to add it and also removed the deprecated `SRC_URI[md5sum]` which should not be added on new upgraded recipes anymore.

Let's see if we pass the CI test again :)

ejoerns (approved):
Now also approved by CI :+1:

PR #215, opened 2022-02-02
Ejo/cherrypick rhu v1.1

einoj:
Support for rauc-hawkbit-updater version 1.1 for dunfell

### DIFF
--- a/recipes-support/rauc-hawkbit-updater/rauc-hawkbit-updater.inc
+++ b/recipes-support/rauc-hawkbit-updater/rauc-hawkbit-updater.inc
@@ -1,0 +1,25 @@
+SUMMARY = "The RAUC hawkBit updater operates as an interface between the RAUC D-Bus API and the hawkBit DDI API."
+HOMEPAGE = "https://github.com/rauc/rauc-hawkbit-updater"
+
+LICENSE = "LGPLv2.1"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=1a6d268fd218675ffea8be556788b780"
+
+inherit cmake pkgconfig systemd useradd
+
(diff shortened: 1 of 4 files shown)


ejoerns:
Thanks for the update, however, I would prefer backporting the patches instead of taking the original ones and adapting them in a follow-up commit.
I am not that happy with backporting things for another 2 years and probably we need to switch to the new override syntax here too at some point. But as long as poky doesn't I would still like to avoid breaking things for users that 'forgot' to update poky..

I also did backports explicitly by mentioning the original commit id and adding a signed-off-by for the backport so far, which I would like to keep.
As there are more things to add and, for the sake of completeness, I attempt to pick them in the order they appeared in master (if possible).

I've done this now and pushed the changes to my test branch which made it through the CI
https://github.com/ejoerns/meta-rauc/commits/dunfell

Thus if there are no objections on this, I would push this into `dunfell` branch quite soon.

ejoerns:
Just noticed I omitted the dt-utils 2021.03 update. I've added this too now.
If CI passes again I'll gonna update 'dunfell'

einoj:
> Thanks for the update, however, I would prefer backporting the patches instead of taking the original ones and adapting them in a follow-up commit. I am not that happy with backporting things for another 2 years and probably we need to switch to the new override syntax here too at some point. But as long as poky doesn't I would still like to avoid breaking things for users that 'forgot' to update poky..
> 
> I also did backports explicitly by mentioning the original commit id and adding a signed-off-by for the backport so far, which I would like to keep. As there are more things to add and, for the sake of completeness, I attempt to pick them in the order they appeared in master (if possible).
> 
> I've done this now and pushed the changes to my test branch which made it through the CI https://github.com/ejoerns/meta-rauc/commits/dunfell
> 
> Thus if there are no objections on this, I would push this into `dunfell` branch quite soon.

I can build your dunfell branch, so that's fine by me

ejoerns:
Thanks for testing! CI looks fine, too. Pushed.

PR #217, opened 2022-02-22
conf: layer: add 'kirkstone' to LAYERSERIES_COMPAT

matthiasklein:
Remove 'honister' that will will be maintained in its own branch.

Signed-off-by: Matthias Klein <matthias@extraklein.de>


ejoerns (approved):
Thanks for bumping the release version right as I saw the autobuilder fail! :wink:

PR #219, opened 2022-03-18
Update RAUC git version to latest master

ejoerns:
Two major new features have found their way into RAUC since v1.6:

* native HTTP(S) streaming support
* bundle encryption

Update the git recipes to optionally allow building the current master.

Introduce convenience variable to enable development version, e.g. from local.conf:

    RAUC_USE_DEVEL_VERSION = "1"


recipes-core/rauc/rauc-git.inc
```diff
@@ -6,9 +6,13 @@ SRC_URI = " \
...
 S = "${WORKDIR}/git"
 
-SRCREV = "a1142006f01291b29bd5159edead178cb3be8ae8"
+SRCREV = "080ad9e75ee4fa36a6093dac29e3d12fec87a167"
 
 DEFAULT_PREFERENCE = "-1"
+
+PACKAGECONFIG[streaming] = "--enable-streaming,--enable-streaming=no,libnl"
+
+PACKAGECONFIG:class-target ??= "service network streaming json nocreate gpt"
```
jluebbe, line 20: How does this interact with
```
PACKAGECONFIG ??= "service network json nocreate gpt"
```
in rauc-target.inc?

recipes-core/rauc/rauc-git.inc
```diff
@@ -11,7 +11,9 @@ S = "${WORKDIR}/git"
 
 SRCREV = "080ad9e75ee4fa36a6093dac29e3d12fec87a167"
 
-DEFAULT_PREFERENCE = "-1"
+RAUC_USE_DEVEL_VERSION[doc] = "Global switch to enable RAUC development (git) version."
```
jluebbe, line 14: Perhaps a comment here how to enable it in `local.conf`?

jluebbe (changes requested):

recipes-core/rauc/rauc-git.inc
```diff
@@ -6,9 +6,15 @@ SRC_URI = " \
...
-PV = "1.1+git${SRCPV}"
+PV = "1.6+git${SRCPV}"
 S = "${WORKDIR}/git"
 
-SRCREV = "a1142006f01291b29bd5159edead178cb3be8ae8"
+SRCREV = "080ad9e75ee4fa36a6093dac29e3d12fec87a167"
 
-DEFAULT_PREFERENCE = "-1"
+RAUC_USE_DEVEL_VERSION[doc] = "Global switch to enable RAUC development (git) version."
+RAUC_USE_DEVEL_VERSION ??= "-1"
```
zyga, line 15: What does -1 imply here? I was expecting a 0 vs 1 flag.

ejoerns: Ma idea was to simply select the same value that `DEFAULT_PREFERENCE` would take. The main purpose of this variable is to allow switching the DEFAULT_PREFERENCE of each RAUC package with a single variable.

For reference: The documentation says about `DEFAULT_PREFERENCE`:
> The most common usage of this is variable is to set it to “-1” within a recipe for a development version of a piece of software.

recipes-core/rauc/rauc-git.inc
ejoerns: Due to my change in 95d041ed56f296847b1cf8eeb89cc203d528cc87, this is now overridden with the new weak default as far as I can see.

recipes-core/rauc/rauc-git.inc
ejoerns: I'd then prefer adding a hint in README. Being able to set global variables in local.conf is not something worth being documented I guess.

recipes-core/rauc/rauc-git.inc
zyga: Ah, I see. Thank you for explaining this.

recipes-core/rauc/rauc-git.inc
ejoerns: I have now added a section in the README

ejoerns:
I had to extend the change in include file handling a bit (see commit messasge).

The update now also comes with a new section in the README and enabling BLK_DEV_NBD in the kernel.

ejoerns:
Updated to latest master commit that contains some further fixes just merged.

ejoerns:
Bumped again. Now to https://github.com/rauc/rauc/commit/3963103bac7dc872c5ee42a2924a39474278af39 (latest master)

ejoerns:
partly cherry-picked to `dunfell` by #243

PR #227, opened 2022-07-07
class: bundle: support incremental image property

jluebbe:
To build bundles with incremental support, we need to define the new property
for relevant images. As RAUC 1.7 already supports (and then ignores) this
property, we can add the support now.


ejoerns (approved):

jluebbe:
@ejoerns: For which branches should we backport this?

PR #229, opened 2022-07-19
class: bundle: support incremental image property (backport to kirkstone)

jluebbe:
This is #227 backported to kirkstone.


ejoerns (commented):
Please add the backport annotation that refers to the original commit (like on the other backported commits).

PR #230, opened 2022-07-19
class: bundle: support incremental image property (backport to dunfell)

jluebbe:
This is #227 backported to dunfell.


ejoerns:
Will not be backported (at least not stand-alone) due to the rename of 'incremental' mode to 'adaptive' mode in RAUC (might go together with #232 once RAUC 1.8 is available in dunfell)

PR #232, opened 2022-08-19
class: bundle: rename 'incremental' image property to 'adaptive'

ejoerns:
While 'incremental' was the first name for the new update optimization mode in RAUC, it will be called 'adaptive' in the RAUC 1.8 release where it is first released with functionality behind.


jluebbe (approved):

ejoerns:
backported to kirkstone

PR #233, opened 2022-08-21
casync: update to latest git version (v2-267-g0efa7ab)

ejoerns:


ejoerns:
FTR: Fixes meta-rauc master build of casync

* `master` branch was renamed to `main`
* poky master provides glibc 2.36 which casync fails building against (fixed in https://github.com/systemd/casync/commit/0efa7abffe5fffbde8c457d3c8fafbdde0bb6e4f)

jluebbe:
I'd prefer to have that information in the commit message.

ejoerns:
Probably worth adding, yes. (However, that it fixes the build was actually not the case when I wrote the commit message initially.)

jluebbe (approved):

ejoerns:
backported to kirkstone

ejoerns:
Backported to 'dunfell'

PR #234, opened 2022-08-22
rauc-git: update SRCREV to current master

jluebbe:


ejoerns (approved):
Looks fine!

FTR: CI build fails because of #233 but rauc and rauc-native already passed.

ejoerns:
backported to kirkstone

PR #237, opened 2022-09-15
dunfell: fix casync branch

om26er:
Well the project apparently now uses `main` which breaks builds for projects that are using casync recipe


ejoerns:
Thanks for pointing this out. Could you please add your Signed-off-by and probably pack the description you have in the PR into the message, too?

An alternative would be to just backport 25806eb37dd78bd2be1b57995fd0a493eeab790c as the bump might contain mainly bug fixes that are useful in dunfell as well.

ejoerns:
I've cherry-picked the corresponding commit from `master` to `dunfell` now.

PR #239, opened 2022-09-30
conf: layer: set LAYERSERIES_COMPAT to 'langdale'

ejoerns:


ejoerns:
Since other PRs cannot be tested properly on master without this and as it is a no-brainer anyway, I'll merge it without further manual review.

PR #240, opened 2022-09-30
rauc: update to 1.8

ejoerns:
Signed-off-by: Enrico Jorns <ejo@pengutronix.de>

PR #242, opened 2022-10-11
rauc-git: update recipe version to 1.8

OnkelUlla:
This has been forgotten in #240.


ejoerns:
No, it has not really been forgotten. The git recipe has always been updated independently of the release recipe since 1.2.
And it is not consistent since the SRCREV points to a commit before 1.8.

Thus updating is good, but with bumping SRCREV, too.

ejoerns (approved):

OnkelUlla:
Yes, you are right - I completely missed the ``SRCREV``. I pushed a corrected version some minutes ago.

PR #248, opened 2022-10-17
classes: bundle: allow skipping bytes from image

a3f:
For most i.MX SoCs, BootROM expects bootloader image to be written into boot media at an offset. While barebox pads the image, so i.MX header starts at that offset, U-Boot doesn't and thus `bundle.bbclass` supported
```
	RAUC_SLOT_bootloader[offset] = "32K"
```
To get U-Boot appropriately padded. Now with the i.MX8M Plus and Nano, the tide turns and the BootROM no longer expects an offset for images installed into eMMC boot partitions. It still expects them for SD and eMMC user area. For U-Boot, this means offset can return to the default of 0, but barebox still has a 32K byte preamble, so images are flashable directly to the start of the SD-Card.

To be able to install barebox into an eMMC boot partition on such systems, we will need to shave off these 32K bytes. Allow doing this with:
```
	RAUC_SLOT_bootloader[offset] = "-32K"
```
Positive offsets will be handled as dd seek as before, while negative will lead to seeks in the input file. I find this more intuitive than the `seek`/`skip` nomenclature with dd.

### DIFF
--- a/classes/bundle.bbclass
+++ b/classes/bundle.bbclass
@@ -42,6 +42,7 @@
 #
 # To prepend an offset to a bootloader image, set the following parameter in bytes.
 # Optionally you can use units allowed by 'dd' e.g. 'K','kB','MB'.
+# If the offset is negative, bytes will not be added, but removed.
 #   RAUC_SLOT_bootloader[offset] ?= "0"
 #
 # Enable building verity format bundles with
@@ -261,8 +262,17 @@ def write_manifest(d):
         if slotflags and 'rename' in slotflags:
             imgname = d.getVarFlag('RAUC_SLOT_%s' % slot, 'rename')
         if slotflags and 'offset' in slotflags:
+            padding = 'seek'
             imgoffset = slotflags.get('offset')
-            if slotflags.get('offset') == '':
+            if imgoffset:
+                sign, magnitude = imgoffset[:1], imgoffset[1:]
+                if sign == '+':
+                    padding = 'seek'
+                    imgoffset = magnitude
+                elif sign == '-':
+                    padding = 'skip'
+                    imgoffset = magnitude
+            if imgoffset == '':
                 imgoffset = '0'
 
         # Keep only the image name in case the image is in a $DEPLOY_DIR_IMAGE subdirectory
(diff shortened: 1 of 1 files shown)


bith3ad:
> > Positive offsets will be handled as dd seek as before, while negative will lead to seeks in the input file. I find this more intuitive than the seek/skip nomenclature with dd.
> 
> I can also add a new var flag like `RAUC_SLOT_bootloader[offset_type]` or something if that's preferred.

That sounds more obvious, so it would be `RAUC_SLOT_bootloader[seek]` and `RAUC_SLOT_bootloader[skip]`

ejoerns:
> I always need to lookup what `[seek]`/`[skip]` means. But we could make `[offset]` an alias for `[seek_offset]` and add a new `[skip_offset]`.
> 
> @ejoerns, What do you prefer?

I would go the simple way and just support the negative offsets. If that actually turns out to be confusing, we could think about a different wording.

In my brain `offset` is the position you move the image around the zero point (which is the partition start).

classes/bundle.bbclass
```diff
@@ -282,8 +291,9 @@ def write_manifest(d):
         if os.path.isfile(searchpath):
             if imgtype == 'boot' and 'offset' in slotflags and imgoffset != '0':
                 subprocess.call(['dd', 'if=%s' % searchpath,
-                                 'of=%s' % bundle_imgpath, 'oflag=seek_bytes',
-                                 'seek=%s' % imgoffset])
+                                 'of=%s' % bundle_imgpath,
+                                 'iflag=skip_bytes', 'oflag=seek_bytes',
+                                 '%s=%s' % (padding, imgoffset)])
```
ejoerns, line 296: `padding` is unset if the `[offset]` varflag argument does not contain a sign.

ejoerns (changes requested):

classes/bundle.bbclass
a3f: Ouch. Fixed.

ejoerns (approved):

a3f:
I am still on Kirkstone for the forseeable time. Is it possible to backport this patch?

ejoerns:
I've cherry-picked this to `kirkstone` to ease cherry-picking #249

PR #249, opened 2022-10-25
classes: bundle: Allow parsing var flags offset, hooks and adaptive

mschwan-phytec:
Allow for BitBake variable parsing inside the variable flags "offset", "hooks" and "adaptive". To enable this, use d.getVarFlag() instead of the builtin get() method.

### DIFF
--- a/classes/bundle.bbclass
+++ b/classes/bundle.bbclass
@@ -263,7 +263,7 @@ def write_manifest(d):
             imgname = d.getVarFlag('RAUC_SLOT_%s' % slot, 'rename')
         if slotflags and 'offset' in slotflags:
             padding = 'seek'
-            imgoffset = slotflags.get('offset')
+            imgoffset = d.getVarFlag('RAUC_SLOT_%s' % slot, 'offset')
             if imgoffset:
                 sign, magnitude = imgoffset[:1], imgoffset[1:]
                 if sign == '+':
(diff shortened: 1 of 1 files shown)


classes/bundle.bbclass
```diff
@@ -281,9 +281,9 @@ def write_manifest(d):
         if slotflags and 'hooks' in slotflags:
             if not have_hookfile:
                 bb.warn("A hook is defined for slot %s, but RAUC_BUNDLE_HOOKS[file] is not defined" % slot)
-            manifest.write("hooks=%s\n" % slotflags.get('hooks'))
+            manifest.write("hooks=%s\n" % d.getVarFlag('RAUC_SLOT_%' % slot, 'hooks'))
```
ejoerns, line 284:
```suggestion
            manifest.write("hooks=%s\n" % d.getVarFlag('RAUC_SLOT_%s' % slot, 'hooks'))
```

classes/bundle.bbclass
```diff
@@ -281,9 +281,9 @@ def write_manifest(d):
...
         if slotflags and 'adaptive' in slotflags:
-            manifest.write("adaptive=%s\n" % slotflags.get('adaptive'))
+            manifest.write("adaptive=%s\n" % d.getVarFlag('RAUC_SLOT_%' % slot, 'adaptive'))
```
ejoerns, line 286:
```suggestion
            manifest.write("adaptive=%s\n" % d.getVarFlag('RAUC_SLOT_%s' % slot, 'adaptive'))
```

ejoerns (changes requested):
Good catch about the parsing handling.

But pretty sure this was only tested for `'offset'` :wink:

mschwan-phytec:
> Good catch about the parsing handling.
> 
> But pretty sure this was only tested for `'offset'` :wink:

Whoops! :grimacing: Fixed.

ejoerns (approved):

mschwan-phytec:
Can we backport this to kirkstone? As adaptive updates are available in kirkstone, too, it would fix the parsing there.

ejoerns:
cherry-picked to `kirkstone`

PR #250, opened 2022-10-28
classes: bundle: warn if RAUC_BUNDLE_FORMAT is not set

ejoerns:
This is the next step towards deprecating the 'plain' format. We start enforcing an explicit decision whether to use old 'plain' or new 'verity' format.

And we guide the user to use 'verity' for new projects.

### DIFF
--- a/classes/bundle.bbclass
+++ b/classes/bundle.bbclass
@@ -201,6 +201,10 @@ def write_manifest(d):
         manifest.write('[bundle]\n')
         manifest.write(d.expand('format=${RAUC_BUNDLE_FORMAT}\n'))
         manifest.write('\n')
+    else:
+        bb.warn('No RAUC_BUNDLE_FORMAT set. This will default to using legacy \'plain\' format.'
+                '\nIf you are unsure, set RAUC_BUNDLE_FORMAT = "verity" for new projects.'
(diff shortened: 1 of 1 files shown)


jluebbe (commented):
Looks good, but perhaps we should point to https://rauc.readthedocs.io/en/latest/reference.html#sec-ref-formats as well?

ejoerns:
Added the link to the documentation

classes/bundle.bbclass
```diff
@@ -201,6 +201,10 @@ def write_manifest(d):
         manifest.write('[bundle]\n')
         manifest.write(d.expand('format=${RAUC_BUNDLE_FORMAT}\n'))
         manifest.write('\n')
+    else:
+        bb.warn('No RAUC_BUNDLE_FORMAT set. This will default to using deprecated \'plain\' format.'
```
jluebbe, line 205: We haven't deprecated 'plain' yet, so perhaps
```suggestion
        bb.warn('No RAUC_BUNDLE_FORMAT set. This will default to using lecacy \'plain\' format.'
```

jluebbe (changes requested):

classes/bundle.bbclass
ejoerns: Ack (for 'legacy')

jluebbe (approved):

ejoerns:
I'd tend to backport this at least to `kirkstone` since I expect this to be out in the field for years. Any objections?

PR #256, opened 2023-01-04
Update layer compat to mickledore

troth:
Upstream oe core has updated the release name from langdale to mickledore.

Signed-off-by: Theodore A. Roth <troth@openavr.org>

### DIFF
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -17,7 +17,7 @@ LAYERRECOMMENDS_rauc = "meta-python"
 # meta-filesystems is needed to build cascync with fuse support (the default)
 LAYERRECOMMENDS_rauc += "meta-filesystems"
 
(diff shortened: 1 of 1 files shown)


troth:
The Ci failure was due to the meta-openembedded layer not having the update to mickledore in place. Since the CI test was run, the update has been merged to the master branch of meta-openembbed.

It doesn't look like I have permissions to manually trigger a new CI run for this. I suspect that the build will run successfully if re-run.

troth:
Is there any chance this will be merged any time soon? It is currently not possible to use meta-rauc with the master branch of oe-core without this change.

ejoerns (approved):

ejoerns:
@troth Sorry, too many things in parallel. The change passes CI and should not have any other impact than making meta-rauc work with master (only), thus ok from my side. Thanks for bumping the compact!

PR #259, opened 2023-01-24
Harmonize bundle file permissions

sirhcel:
The casync bundle is no executable file and should have the same permissions as the regular bundle file.


ejoerns (approved):
Looks good!

@sirhcel Thanks for the quick fix!

PR #267, opened 2023-04-02
Update to RAUC v1.9

OnkelUlla:


ejoerns (approved):
It's not a simple cherry-pick as stated in the comment, but since only minor adaptions were made to reflect the slightly different include hierarchy, I think it's okay like it is.

Thanks!

PR #271, opened 2023-06-06
Add Two Ways of Adding Custom Meta-Data Or Settings to The Manifest

ejoerns:
With `RAUC_META_SECTIONS` there is a similar approach to how bundles are defined that allows to specify RAUC [meta data sections](https://rauc.readthedocs.io/en/latest/reference.html?highligh=meta#manifest) in the manifest.

With `RAUC_MANIFEST_EXTRA_LINES` any (RAUC-parsable) manifest information can be appended to a manifest.

### DIFF
--- a/classes-recipe/bundle.bbclass
+++ b/classes-recipe/bundle.bbclass
@@ -89,6 +89,11 @@
 #   RAUC_META_mydata[release-notes] = "a few notes here"
 #
 #   RAUC_META_foo[bar] = "baz"
+#
+# Adding any sort of additional lines to the manifest can be done with the
+# RAUC_MANIFEST_EXTRA_LINES variable (using '\n' to indicate newlines):
+#
+#   RAUC_MANIFEST_EXTRA_LINES = "[section]\nkey=value\n"
 
 LICENSE ?= "MIT"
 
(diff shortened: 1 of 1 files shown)


classes-recipe/bundle.bbclass
```diff
@@ -79,6 +79,21 @@
...
 #
 #   RAUC_CASYNC_BUNDLE = "1"
+#
+# To define custom manifest 'meta' sections, you may use
+# 'RAUC_META_SECTIONS' as follows:
+#
+#   RAUC_META_SECTIONS = "mydata foo"
+#
+#   RAUC_META_mydata[release-type] = "beta"
+#   RAUC_META_mydata[release-note] = "a few words here"
```
jluebbe, line 89: Also in the commit message:
```suggestion
#   RAUC_META_mydata[release-notes] = "a few words here"
```

jluebbe (changes requested):
Looks good otherwise.

classes-recipe/bundle.bbclass
ejoerns: Fixed :+1:

jluebbe (approved):

BenediktCyranka:
Works perfect

ejoerns:
cherry-picked to `kirkstone`

PR #276, opened 2023-07-05
README.rst: add Contribution notes, add DCO, add backporting notes

ejoerns:


DCO
```diff
@@ -0,0 +1,39 @@
+Developer's Certificate of Origin
+---------------------------------
+
+meta-rauc uses the `Developer's Certificate of Origin 1.1
+<https://developercertificate.org/>`_ with the same `process
+<https://www.kernel.org/doc/html/latest/process/submitting-patches.html#sign-your-work-the-developer-s-certificate-of-origin>`_
+as used for the Linux kernel:
```
OnkelUlla, line 7: That [process](https://www.kernel.org/doc/html/latest/process/submitting-patches.html#sign-your-work-the-developer-s-certificate-of-origin) evolved a little bit in the meantime in kernel versions [v5.10](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=9f364b605f34e1c0b9847943f55db061d5a26bcb), [v5.11](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=7d71788735608b1578e3615ff2646d3a089ff0a8) and [v6.3](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=d4563201f33a022fc0353033d9dfeb1606a88330) - do we want to follow?

README.rst
```diff
@@ -173,8 +173,21 @@ in your local.conf. Note that this has the same effect as setting
...
+=================
+
+To report bugs, file a new `issue <https://github.com/rauc/meta-rauc/issues>`_
+on GitHub.
+
+For fixing bugs, bumping recipes or adding new features, open a `pull request
+<https://github.com/rauc/meta-rauc/pulls>`_ on GitHub.
+
+Add a ``Signed-off-by`` line to your commits according to the
+`Developer’s Certificate of Origin
```
OnkelUlla, line 186: See my first nitpicking note regarding the DCO.

README.rst
```diff
@@ -186,6 +186,24 @@ Add a ``Signed-off-by`` line to your commits according to the
 `Developer’s Certificate of Origin
 <https://github.com/rauc/meta-rauc/blob/master/DCO>`_.
 
+Backporting
+-----------
+
+For backporting changes to a stable or LTS branch, two options exist:
+
+a) drop a backport request in the original pull request
+b) backport on your own an create a new pull request
```
OnkelUlla, line 195:
```suggestion
b) backport on your own and create a new pull request
```

OnkelUlla (changes requested):
At least the typo should be fixed. If we discuss following the evolution of the kernel's process of submitting patches is up to you.

DCO
ejoerns: I am unsure what affects us here. Most changes refer to git workflow over mail. The only change I see is that they seemed to have allowed contributing by pseudonym. Am I right?

Maybe we should not write 'same' process here anyway.

jluebbe: Most changes did not affect the DCO, and we point only to the "Sign your work" section. Furthermore, we already accepted changes under pseudonym, so I don't see any need to diverge from the kernel here.

README.rst
ejoerns: Fixed this

DCO
OnkelUlla: > Ok, just noted the other changes are in chapters we do not reference. Thus the difference is mentioning of `git commit -s` (which we already had), mentioning of `git revert -s`, and allowing pseudonyms.

That's correct.
And as already stated before: I don't really care if we follow that evolution or not, but if we don't then we should not talk of the _same_ process as found in the kernel.

ejoerns: I have no strong opinion on pseudonyms. I prefer actual names but if I have to choose between pseudonym or no contribution I would take the pseudonym.

As it is the same for RAUC itself: @jluebbe Do you have an opinion? Otherwise we could just take the text from the kernel 1:1 and continue speaking about 'same'.

ejoerns:
Is anyone in the mood for a review? :smirk:

PR #289, opened 2023-11-08
fix: bundle.bbclass: varflags expansion

Jarsop:
~Python anonymous functions are evaluated before `=` which expand the final variable value. This fix set the `getVarFlags` kwarg `expand` to `True` (default is `False`) to be able to assign varflag with bitbake variable (`${...}`) or python inline syntax (`${@...})`.~

Python anonymous functions are evaluated before the `=` sign and Python tasks do not expand variables. This fix introduces a set of varflags via `RAUC_VARFLAGS_SLOTS` and `RAUC_VARFLAGS_HOOKS` variables.
It sets the `getVarFlags` keyword argument `expand` to the respective list (default is `False`). This allows for assigning varflags using bitbake variables (`${...}`) or Python inline syntax (`${@...}`).

### DIFF
--- a/classes-recipe/bundle.bbclass
+++ b/classes-recipe/bundle.bbclass
@@ -138,11 +138,15 @@ RAUC_CASYNC_BUNDLE ??= "0"
 RAUC_BUNDLE_FORMAT ??= ""
 RAUC_BUNDLE_FORMAT[doc] = "Specifies the bundle format to be used (plain/verity)."
 
+RAUC_VARFLAGS_SLOTS = "name type fstype file hooks adaptive rename offset depends"
+RAUC_VARFLAGS_HOOKS = "file hooks"
+
 # Create dependency list from images
 python __anonymous() {
     d.appendVarFlag('do_unpack', 'vardeps', ' RAUC_BUNDLE_HOOKS')
     for slot in (d.getVar('RAUC_BUNDLE_SLOTS') or "").split():
-        slotflags = d.getVarFlags('RAUC_SLOT_%s' % slot)
+        slot_varflags = d.getVar('RAUC_VARFLAGS_SLOTS').split()
+        slotflags = d.getVarFlags('RAUC_SLOT_%s' % slot, expand=slot_varflags)
         imgtype = slotflags.get('type') if slotflags else None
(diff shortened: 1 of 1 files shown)


Jarsop:
Same remark with ci commit, see https://github.com/rauc/meta-rauc/pull/290#issuecomment-1802913597

Jarsop:
@ejoerns Second commit introduce a refactoring as it's dependent from the first but it can be push through another branch if needed.
BTW have we a test bundle recipes ? If no maybe can I create a dedicated test meta layer to have some test recipes to test our different classes ? (And sorry for the notifications noise but I had an issue with my git history...)

ejoerns:
@Jarsop I really appreciate your changes! Especially the dict handling simplification looks really useful!

I have slightly modified the commit title to better match our conventions (removed conventional commit prefix).

> @ejoerns Second commit introduce a refactoring as it's dependent from the first but it can be push through another branch if needed.

Should not be necessary, thx.

> BTW have we a test bundle recipes ? If no maybe can I create a dedicated test meta layer to have some test recipes to test our different classes ? (And sorry for the notifications noise but I had an issue with my git history...)

No, we currently don't have but actually should. So feel free to make a proposal!

ejoerns (approved):

Jarsop:
> @Jarsop I really appreciate your changes! Especially the dict handling simplification looks really useful!
> 
> I have slightly modified the commit title to better match our conventions (removed conventional commit prefix).

You're welcome and thanks for the realignment.

> No, we currently don't have but actually should. So feel free to make a proposal!

Ok, I will think about that and will make a proposal, maybe fake images, parsing around `bitbake -e` etc.

Jarsop:
@ejoerns  can we consider using [`kas`](https://github.com/siemens/kas) to configuring and build recipes/test images ?

ejoerns:
> @ejoerns can we consider using [`kas`](https://github.com/siemens/kas) to configure and build recipes/test images ?

I'd prefer using new built-in layer tooling, but if kas actually gives a benefit there, I would not strictly be against it.

ejoerns:
backported to `kirkstone`.

PR #295, opened 2023-12-06
classes-recipe: bundle: use native paths instead of target paths

mqqc:
Any recipe using the `bundle` bbclass will fail if either:
- `${prefix}` is not equal to `${prefix_native}`, or:
- `${prefix_native}` is not equal to `/usr`, or:
- `${bindir}` is not equal to `${bindir_native}`

This PR corrects this by using `${prefix_native}` and `${STAGING_BINDIR_NATIVE}` where appropriate.


ejoerns (approved):
Indeed, this has always been wrong. Thanks for fixing! :+1:

PR #307, opened 2024-03-16
rauc: rauc-mark-good: disable implicit dependencies

js731ca:
Having implicit dependencies enabled would lead to ordering cycles when enabling 'boot assessment', through
systemd-boot-check-no-failures [1], which is a service that forces the multi-user.target to be reached before the boot-complete.target.

The ordering cycle in this case would be:
boot-complete -requires-> systemd-boot-check-no-failures systemd-boot-check-no-failures -after-> multi-user rauc-mark-good -after-> boot-complete
boot-complete -after-> multi-user
multi-user -after-> rauc-mark-good

circle

Which systemd would break by simply not running rauc-mark-good.

The solution is to disable implicit dependencies that would otherwise add an 'After=rauc-mark-good.service' to the multi-user.target via the [Install] section.

Link: https://www.freedesktop.org/software/systemd/man/latest/systemd-boot-check-no-failures.html

### DIFF
--- a/recipes-core/rauc/files/rauc-mark-good.service
+++ b/recipes-core/rauc/files/rauc-mark-good.service
@@ -4,6 +4,7 @@ ConditionKernelCommandLine=|bootchooser.active
 ConditionKernelCommandLine=|rauc.slot
 After=boot-complete.target
 Requires=boot-complete.target
+DefaultDependencies=no
 
 [Service]
 Type=oneshot


ejoerns:
> Having implicit dependencies enabled would lead to ordering cycles when enabling 'boot assessment', through
systemd-boot-check-no-failures [1], which is a service that forces the multi-user.target to be reached before the boot-complete.target.

Even if not using the example `systemd-boot-check-no-failures` directly, it is probably a good blueprint to order `multi-user.target` before `boot-complete.target`, yes.

@js731ca thank you for diving into this.

> The solution is to disable implicit dependencies that would otherwise add an 'After=rauc-mark-good.service' to the multi-user.target via the [Install] section.

I had to look this up, so for reference: this addresses systemd behavior that's described in https://www.freedesktop.org/software/systemd/man/latest/systemd.target.html#Default%20Dependencies

Setting `DefaultDependencies=no` sounds reasonable to me. An alternative would be not to use `WantedBy=multi-user.target` for installation. But I actually don't know a better option, yet.

@michaelolbrich maybe?

michaelolbrich:
This looks like the correct solution to me. And the equivalent service from systemd (`systemd-bless-boot.service`) contains `DefaultDependencies=no` as well.

ejoerns (approved):
@js731ca @michaelolbrich thanks for your opinions on this!

So we go for the `DefaultDependencies=no` solution.

FTR: CI fails due to the PR sitting on older master but the change itself cannot make CI fail, anyway.

PR #308, opened 2024-03-17
layer.conf: switch to scarthgap series

troth:
Since oe-core switched to `scarthgap` release series only (i.e. removed `nanbield` compatiblity), we have to update our `layer.conf` file.

### DIFF
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -17,7 +17,7 @@ LAYERRECOMMENDS_rauc = "meta-python"
 # meta-filesystems is needed to build cascync with fuse support (the default)
 LAYERRECOMMENDS_rauc += "meta-filesystems"
 
-LAYERSERIES_COMPAT_rauc = "nanbield"
+LAYERSERIES_COMPAT_rauc = "nanbield scarthgap"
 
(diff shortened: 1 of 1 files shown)


conf/layer.conf
```diff
@@ -17,7 +17,7 @@ LAYERRECOMMENDS_rauc = "meta-python"
 # meta-filesystems is needed to build cascync with fuse support (the default)
 LAYERRECOMMENDS_rauc += "meta-filesystems"
 
-LAYERSERIES_COMPAT_rauc = "nanbield"
+LAYERSERIES_COMPAT_rauc = "scarthgap"
```
zandrey, line 20: According to [_Yocto Project Releases_](https://wiki.yoctoproject.org/wiki/Releases), `nanbield` is still supported until May 2024, so I would not drop the compatibility here.

troth: Updated commit to not drop `nanbield`.

ejoerns: We have a dedicated `nanbield` branch for `nanbield` anyway, but I'm fine with keeping compatibility in master first of all, too.

ejoerns (approved):

Issue #299, opened 2024-01-15
When IMAGE_NAME_SUFFIX is non standard a bundle will "fail to find sources"


Gerry546:
Currently `bundle.bbclass` assumes that [IMAGE_NAME_SUFFIX ](https://docs.yoctoproject.org/ref-manual/variables.html?highlight=extra_#term-IMAGE_NAME_SUFFIX) is using the default value of  `.rootfs`.
If you change IMAGE_NAME_SUFFIX to something else an error will be reported when building the bundle.

ejoerns:
@Gerry546 You're right, I've created #300 to address this. Thanks!

PR #312, opened 2024-04-08
bundle.bbclass: use IMAGE_MACHINE_SUFFIX instead of hard-coded '-${MACHINE}'

jdelbefoss:
Similar to #299, when changing the default IMAGE_MACHINE_SUFFIX the bundle build fails (error in do_configure()).
The fix is identical to #300.


ejoerns:
FTR: `IMAGE_MACHINE_SUFFIX` was introduced in https://git.yoctoproject.org/poky/commit/?id=c0df9c9e92b56c33952fa3e6c050f1e4c53b66d9 and is only available since mickledore. Thus no LTS backporting is required.

ejoerns (approved):
@jdelbefoss Thank you for fixing!

PR #320, opened 2024-05-26
Follow WORKDIR -> UNPACKDIR transition

ejoerns:
This adapts to the oe-core rework to enforce a separate directory for unpacking local sources (`UNPACKDIR`) instead of polluting `WORKDIR` directly.

Follows the preliminary guidelines from:
https://lists.openembedded.org/g/openembedded-architecture/message/2007

For this, we need to break compatibility with all releases before `master`.

### DIFF
--- a/classes-recipe/bundle.bbclass
+++ b/classes-recipe/bundle.bbclass
@@ -51,7 +51,7 @@
 #
 # To add additional artifacts to the bundle you can use RAUC_BUNDLE_EXTRA_FILES
 # and RAUC_BUNDLE_EXTRA_DEPENDS.
-# For files from the WORKDIR (fetched using SRC_URI) you can write:
+# For files from the UNPACKDIR (fetched using SRC_URI) you can write:
 #
 #   SRC_URI += "file://myfile"
 #   RAUC_BUNDLE_EXTRA_FILES += "myfile"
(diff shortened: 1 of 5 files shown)


troth:
Testing your patches for this and I see the following WARNING:
```
WARNING: rauc-conf-1.0-r0 do_unpack: rauc-conf: the directory ${WORKDIR}/${BP} (/home/troth/work/judo/judo-main/build/tmp/work/armv7at2hf-neon-trimble-linux-gnueabi/rauc-conf/1.0/rauc-conf-1.0) pointed to by the S variable doesn't exist - please set S within the recipe to point to where the source has been unpacked to
```

Adding the following after `SRC_URI` should quell the warning:

```
S = "${WORKDIR}/sources"
UNPACKDIR = "${S}"
```

ejoerns:
@troth Indeed, I get the same error when building rauc-conf. No idea why I missed this, thanks for the hint!

I also wasn't fully aware yet that setting S became mandatory with the recent oe-core changes.

ejoerns:
Should have read this paragraph :)

> Previously, S was always created but after the recent changes it is
> not. This means the check in do_unpack_qa triggers where S is not
> created by a recipe where as it did not before. This can mean adding a
> S definition to a recipe that only uses file:// SRC_URI entries.

Emantor (approved):

PR #324, opened 2024-06-07
bundle.bbclass: do not generate SSTATE artifacts

ejoerns:
The artifacts created are quite large and keeping multiple of them will consume a notable amount of disk space while the expected speed-up is minimal.
(For the same reasons, the image class does not generate SSTATE artifacts).

Having SSTATE artifacts for bundles enabled can also lead to unexpected/unwanted behavior if a rootfs image recipe was rebuild and the binary output (file system image) changed but did not make it into the final bundle.
This becomes a major problem, e.g. in the context of verified boot.

### DIFF
--- a/classes-recipe/bundle.bbclass
+++ b/classes-recipe/bundle.bbclass
@@ -454,6 +454,8 @@ addtask bundle after do_configure
 
 inherit deploy
 
+SSTATE_SKIP_CREATION:task-deploy = '1'
+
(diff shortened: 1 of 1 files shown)


Emantor (approved):

ejoerns:
backported to `scarthgap`

ejoerns:
backported to `kirkstone`

PR #327, opened 2024-06-18
README: rauc-conf needs to be appended now instead of rauc

ejoerns:
With d51bab8f (rauc: split package into 'rauc' for binary and 'rauc-conf' for configuration), users need to append rauc-conf instead of rauc when providing their device-specific configuration.

This was not reflected by the docs yet.

Also add an explicit migration note early in the README.rst.

### DIFF
--- a/scripts/README
+++ b/scripts/README
@@ -17,8 +17,8 @@ To use them in your BSP in order to sign and verify bundles, you have to:
 
    Copy the keyring file `dev/ca.cert.pem` into your projects meta layer under
    `recipes-core/rauc/files` and add a simple rauc .bbappend file
-   `recipes-core/rauc/rauc_%.bbappend` that adds the keyring file to the rauc
-   package search path:
+   `recipes-core/rauc/rauc-conf.bbappend` that adds the keyring file to the rauc
+   config package search path:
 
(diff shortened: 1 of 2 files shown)


jluebbe (approved):

scripts/README
```diff
@@ -17,8 +17,8 @@ To use them in your BSP in order to sign and verify bundles, you have to:
 
    Copy the keyring file `dev/ca.cert.pem` into your projects meta layer under
    `recipes-core/rauc/files` and add a simple rauc .bbappend file
```
OnkelUlla, line 19: Formally in line 19 this should be _"and add a simple rauc-conf .bbappend file"_, perhaps it would be better to just drop the first part and use _"and add a simple .bbappend file"_ instead.

OnkelUlla (changes requested):
Looks good otherwise.

scripts/README
ejoerns: :+1:  I would be fine with both variants. Since the PR was merged a minute before your comment, feel free to open a new one with this additional fix.

OnkelUlla: Done and already merged.

ejoerns:
backported to `scarthgap`

PR #330, opened 2024-07-01
rauc-git: update to latest SRCREV

ejoerns:
With the latest master branch state, RAUC gained support for installing a systemd catalog file.
Add the catalog path to the common `FILES:${PN}-service` for `rauc-target.inc` sine it will not have an impact when using it with a release not containing the catalog file.


jluebbe (approved):

ejoerns:
backported to `scartghap`

PR #333, opened 2024-07-23
curl: enable PACKAGECONFIG for HTTP/2 (nghttp2) on the target

jluebbe:
Without HTTP/2 support, streaming installation is significantly slower.

### DIFF
--- a/recipes-support/curl/curl_%.bbappend
+++ b/recipes-support/curl/curl_%.bbappend
@@ -1,0 +1,1 @@
+require ${@bb.utils.contains('DISTRO_FEATURES', 'rauc', '${BPN}_rauc.inc', '', d)}
--- a/recipes-support/curl/curl_rauc.inc
+++ b/recipes-support/curl/curl_rauc.inc
@@ -1,0 +1,5 @@
+# HTTP/2 allows having multiple parallel (range-) requests in flight on the
+# same connection without requiring multiple TCP and TLS handshakes.
+# This speeds up adaptive updates a lot, since they work by making many
+# range-requests for parts of the update bundle.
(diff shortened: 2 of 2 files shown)


recipes-support/curl/curl_%.bbappend
```diff
@@ -0,0 +1 @@
+require ${@bb.utils.contains('DISTRO_FEATURES', 'rauc', '${BPN}_rauc.inc', '', d)}
```
hnez, line 1: A comment about the benefits of HTTP/2 for adaptive updates would be nice and why it justifies manipulating the PACKAGECONFIG of such an essential recipe.

Something like:

```bash
# HTTP/2 allows having multiple parallel (range-) requests in flight on the same
# connection without requiring multiple TCP and TLS handshakes.
# This speeds up adaptive updates a lot, since they work by making a lot of
# range-requests for parts of the update bundle.
```

hnez (approved):
The change looks good and justified to me, since the speedups achievable via HTTP/2 in the adaptive update case are so large.
As mentioned a comment would be nice, so that one does not have to refer to the commit message.

I am currently building a bundle based on this and will check if HTTP/2 works with it.
Since curl is used in a lot of packages this takes some time.

hnez:
I've tested the change now and with it applied I see `GET /lxatac-core-bundle-base-lxatac.raucb HTTP/2.0` access log lines on my HTTP/2 capable webserver.

recipes-support/curl/curl_%.bbappend
jluebbe: I've added the comment.

ejoerns:
backported to `scarthgap`

PR #337, opened 2024-08-07
bundle.bbclass: support manifest generation for artifacts

jluebbe:
This is useful only in combination with the upcoming support for artifacts, but shouldn't hurt the standard case for now.

### DIFF
--- a/classes-recipe/bundle.bbclass
+++ b/classes-recipe/bundle.bbclass
@@ -406,7 +406,7 @@ CASYNC_BUNDLE_LINK_NAME ??= "${CASYNC_BUNDLE_BASENAME}-${MACHINE}"
 CASYNC_BUNDLE_EXTENSION ??= "${BUNDLE_EXTENSION}"
 CASYNC_BUNDLE_EXTENSION[doc] = "Specifies desired custom filename extension of generated RAUC casync bundle."
 
-do_bundle() {
+fakeroot do_bundle() {
 	if [ -z "${RAUC_KEY_FILE}" ]; then
 		bbfatal "'RAUC_KEY_FILE' not set. Please set to a valid key file location."
 	fi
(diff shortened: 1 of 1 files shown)


classes-recipe/bundle.bbclass
```diff
@@ -40,6 +40,12 @@
 #   RAUC_SLOT_rootfs ?= "core-image-minimal"
 #   RAUC_SLOT_rootfs[rename] ?= "rootfs.ext4"
 #
+# To generate an artifact image, use <repo>/<artifact> as the image name:
+#   RAUC_BUNDLE_SLOTS += "containers/test"
+#   RAUC_SLOT_containers/test ?= "container-test-image"
+#   RAUC_SLOT_containers/test[fstype] = "tar.gz"
+#   RAUC_SLOT_containers/test[convert] = "extract;composefs"
```
ejoerns, line 47: Maybe list existing features here first of all (i.e. ``tar-extract``)

classes-recipe/bundle.bbclass
```diff
@@ -291,6 +297,8 @@ def write_manifest(d):
             manifest.write("hooks=%s\n" % slotflags.get('hooks'))
         if 'adaptive' in slotflags:
             manifest.write("adaptive=%s\n" % slotflags.get('adaptive'))
+        if 'convert' in slotflags:
+            manifest.write("convert=%s\n" % slotflags.get('convert'))
```
ejoerns, line 301: Doesn't this require to run `do_bundle` as fakeroot task, too?

classes-recipe/bundle.bbclass
```diff
@@ -40,6 +40,12 @@
 #   RAUC_SLOT_rootfs ?= "core-image-minimal"
 #   RAUC_SLOT_rootfs[rename] ?= "rootfs.ext4"
 #
+# To generate an artifact image, use <repo>/<artifact> as the image name:
```
ejoerns, line 43: I'm not quite happy with the syntax and the usage of 'SLOT' for artifacts, but might be the easiest solution first.

classes-recipe/bundle.bbclass
jluebbe: Probably, but that needs some more testing. This can't break existing use-cases though.

classes-recipe/bundle.bbclass
jluebbe: `SLOT` is wrong here anyway (it should be `IMAGE`), but changing that in a backwards compatible way will be tricky.

ejoerns:
I've added the changes that I needed to make this work. I don't think it makes much sense to merge this without having the pseudo/fakeroot setup working correctly. Except for the documentation, the changes are only related to *conversion* for adaptive updates anyway.

What would still be needed probably is to remove the now potentially useless pseudo handling for casync and check for potential breakage of use cases due to the added `fakeroot` to `do_bundle()`.

ejoerns:
Rebased onto master to send this through the CI workflow.

ejoerns:
> I've added the changes that I needed to make this work. I don't think it makes much sense to merge this without having the pseudo/fakeroot setup working correctly. Except for the documentation, the changes are only related to _conversion_ for adaptive updates anyway.
> 
> What would still be needed probably is to remove the now potentially useless pseudo handling for casync and check for potential breakage of use cases due to the added `fakeroot` to `do_bundle()`.

This is fixed now by https://github.com/rauc/rauc/pull/1603 and I've added a seperate commit to enable pseudo and drop the old pseudo/fakeroot handling for casync.

jluebbe (commented):
Thanks, looks good to me now.

ejoerns (approved):

PR #336, opened 2024-07-24
bundle.bbclass: fix variable expansion of RAUC_META_* varflags

FederAndInk:
variables assigned to `RAUC_META_<section>[varflag]` are not expanded anymore.

This is caused by `getVarFlags` not expanding variables by default, revert to using `getVarFlag` which does expand variables by default.

fix #335

partially revert changes introduced in ffba61f2f79ec03395629d8713a59226b0158d5f

PR #343, opened 2024-08-22
classes: bundle: use getVarFlag instead of .get

js731ca:
Make use of .getVarFlag instead of the simple .get to allow bitbake level variable expansion.

This is useful especially when using https://rauc.readthedocs.io/en/latest/advanced.html#sec-variants with the device-tree compatible constructed from another variable in e.g. the machine config

for consistencies sake the other/close `slotfllag.get` calls where also replaced

---
Example, from a bundle recipe:
```
RAUC_SLOT_bootloader = "barebox-foo"
RAUC_SLOT_bootloader[name] = "bootloader.foo,product-Rev${MACHINE_REVISION}"                                      
RAUC_SLOT_bootloader[type] = "boot"
...
```

---
Sidenote: i currently don't have a full build+config setup for something newer than `kirkstone`... but the changes should port easily?

### DIFF
--- a/classes/bundle.bbclass
+++ b/classes/bundle.bbclass
@@ -231,17 +231,17 @@ def write_manifest(d):
     for slot in (d.getVar('RAUC_BUNDLE_SLOTS') or "").split():
         slotflags = d.getVarFlags('RAUC_SLOT_%s' % slot)
         if slotflags and 'name' in slotflags:
-            slotname = slotflags.get('name')
+            slotname = d.getVarFlag('RAUC_SLOT_%s' % slot, 'name')
         else:
             slotname = slot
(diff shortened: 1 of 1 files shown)


jluebbe:
This part has been changed significantly by #289 and #336 since kirkstone. Variable expansion should work there.

@ejoerns Should those changes be backported to kirkstone or would you take this change?

ejoerns:
@js731ca Thank your for your fix. As @jluebbe already mentioned, we have existing fixes for this in `master`/`scarthgap`.

I just checked that they are backportable with manageable effort and seem to work well under `kirkstone`, too.

I prefer the backport way since it ensures that `master` and `kirkstone` version of `bundle.bbclass` stay more similar which should ease further backports.

PR #344, opened 2024-08-23
use pengutronix-hosted runner

jluebbe:
This is based on @hnez's [forrest runner](https://github.com/hnez/forrest), which provides ephemeral VMs while still allowing efficient and secure access to download and sstate caches.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,9 +23,9 @@ jobs:
         with:
           path: meta-rauc
       - name: Clone poky
-        run: git clone -b master git://git.yoctoproject.org/poky
+        run: git clone --shared --reference-if-able /srv/shared-git/poky.git -b master https://github.com/yoctoproject/poky.git
       - name: Clone meta-openembedded
-        run: git clone -b master https://github.com/openembedded/meta-openembedded.git
+        run: git clone --shared --reference-if-able /srv/shared-git/meta-openembedded.git -b master https://github.com/openembedded/meta-openembedded.git
       - name: Initialize build directory
         run: |
           source poky/oe-init-build-env build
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,81 @@
+name: build
+
+on:
+  push: {}
+  pull_request: {}
+  # allow rebuilding without a push
+  workflow_dispatch: {}
```
hnez, line 7: I would suggest adding a weekly or monthly [`on: schedule:`](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule) run to keep the sstate fresh and to check if everything still works with the current master of all the other layers.

But that's an unrelated change that could be done in a separate PR.

hnez (approved):

ejoerns (approved):

.github/workflows/build.yml
```diff
@@ -0,0 +1,81 @@
...
+    name: meta-rauc Build
+    # run on self-hosted runner for the main repo or if vars.BUILD_RUNS_ON is set
+    runs-on: >-
+      ${{
+        (vars.BUILD_RUNS_ON != ''  && fromJSON(vars.BUILD_RUNS_ON)) ||
+        (github.repository == 'rauc/meta-rauc' && fromJSON('["self-hosted", "forrest", "build"]')) ||
+        'ubuntu-20.04'
+      }}
+    # abort if it seems that we're rebuilding too much
+    timeout-minutes: 120
```
ejoerns, line 20: Won't we get stuck here then? Currently, we don't save the sstate cache on errors/timeouts, right?

jluebbe: A full rebuild on our runner needs less than 2 hours.

ejoerns: Sounds promising.

PR #345, opened 2024-08-27
.github/workflows/build: pre-build sstate regularly

jluebbe:


hnez (approved):
Looks good! We will have to see if running this daily turns out to be excessive.
In that case we can reduce it to once or multiple times a week.
But as long as nothing relevant changes upstream these build should be very fast.

ejoerns (approved):

ejoerns:
"Successful in 3m" sounds great btw.

PR #350, opened 2024-11-14
rauc: prepare support for locating the config in /usr or /run

a3f:
RAUC PR #1557[1] extends RAUC to look up its system.conf in /etc/rauc, /run/rauc or /usr/lib/rauc in that order.

The RAUC recipe understandably hardcodes the assumption that the system.conf must be supplied and that it must be located in /etc/rauc.

To make the recipe more useful with RAUC once #1557 is merged, let's do two tiny adjustments:

  - Add RAUC_CONF_DIR, which default to /etc/rauc, but can be set by the user to be /usr/lib/rauc instead for hermetic-usr use cases.

  - Add RAUC_CONF_INSTALL, which allows skipping installation of the config altogether. This is useful if the RAUC config is always dynamically generated in /run/rauc and having an always masked /usr/lib/rauc/system.conf would just be confusing.

To avoid breaking existing users, the defaults are chosen, so the system.conf continues to be installed in /etc. This PR does not break anything for current RAUC versions, so it should be fine to merge it independent of the update to a future RAUC release that integrates the upstream PR.

[1]: https://github.com/rauc/rauc/pull/1557


a3f:
This PR can't be trivially rebased onto master and my BSP still uses Kirkstone. To avoid churn, I'd like to reach agreement first whether this approach is ok before rebasing.

recipes-core/rauc/files/system.conf
```diff
@@ -1,6 +1,6 @@
-## This is an example RAUC system configuration. This file will be installed
-## into /etc/rauc/system.conf on your target and describes your system from the
-## perspective of the RAUC update service.
+## This is an example RAUC system configuration. This file will be be looked up
+## in {/etc,/run,/usr/lib}/rauc/system.conf on your target and describes your
+## system from the # perspective of the RAUC update service.
```
jluebbe, line 3:
```suggestion
## system from the perspective of the RAUC update service.
```

README.rst
```diff
@@ -88,9 +88,9 @@ you have to follow at least the following steps:
      DISTRO_FEATURES += "rauc"
 
 2. Add a ``rauc_%.bbappend`` in your device-specific (BSP) layer
-   that installs your RAUC system configuration file under
-   ``/etc/rauc/system.conf``. For information on how to write the RAUC
-   update file, please refer to the RAUC user documentation [1]_::
+   that installs your RAUC system configuration, so RAUC can find it.
```
jluebbe, line 91:
```suggestion
   that installs your RAUC system configuration, so RAUC can find it (in `/etc/rauc/` or `/usr/lib/rauc/`).
```

recipes-core/rauc/rauc-target.inc
```diff
@@ -1,10 +1,12 @@
 RAUC_KEYRING_FILE ??= "ca.cert.pem"
 RAUC_KEYRING_URI ??= "file://${RAUC_KEYRING_FILE}"
+RAUC_CONF_INSTALL ??= "1"
+RAUC_CONF_DIR ??= "${sysconfdir}/rauc"
```
jluebbe, line 4: Couldn't this be just one option? If the user overrides `RAUC_CONF_DIR` to `""`, just don't install the config.

jluebbe (changes requested):

recipes-core/rauc/rauc-target.inc
a3f: IMO,
```
RAUC_CONF_INSTALL = "0"
RAUC_CONF_DIR = "${nonarch_libdir}/rauc"
```

can be useful if the user wants to install `ca.cert.pem` to `/usr/lib`, but wants the config to be in `/run/rauc`. What do you think?

ejoerns (commented):
I am not convinced if it is worth changing this for kirkstone.
I assume no one still sitting on a kirkstone-based BSP will transition to a different config directory.

For `scarthgap` however, it might make sense to support it.
However, since scarthgap we also use a separate [rauc-conf.bb](https://github.com/rauc/meta-rauc/blob/master/recipes-core/rauc/rauc-conf.bb) recipe and it should be quite easy to just write a custom one that does not install the `system.conf`.

For the path my favorite choice would be to just propose `/usr/lib/rauc/system.conf` as the default in `master` and add a note to the migration section.
People then can still use a different path in their custom rauc-conf recipes if required.

If it makes things easier, I would, however, be open to taking the minimalistic `RAUC_CONF` variable approach for both `master` and `scarthgap`.

a3f:
> I am not convinced if it is worth changing this for kirkstone.

It doesn't introduce any functional change on its own, just makes it easier for users to change the default if needed.

> I assume no one still sitting on a kirkstone-based BSP will transition to a different config directory.

I will :-)

> [I]t should be quite easy to just write a custom one that does not install the system.conf.

I think it would still be good to have a variable that can just be overridden to get the old behavior back (if we make it the default) or to opt-in to hermetic-usr.

> For the path my favorite choice would be to just propose /usr/lib/rauc/system.conf as the default in master and add a note to the migration section.

Yes, that sounds sensible for master.

> If it makes things easier, I would, however, be open to taking the minimalistic RAUC_CONF variable approach for both master and scarthgap.

How about we use that minimalistic approach in all of kirkstone, scarthgap and master and additionally change master's default..?

ejoerns:
> > [I]t should be quite easy to just write a custom one that does not install the system.conf.
> 
> I think it would still be good to have a variable that can just be overridden to get the old behavior back (if we make it the default) or to opt-in to hermetic-usr.

My comment was about not installing `system.conf` at all. I guess we won't make this the default.

> > If it makes things easier, I would, however, be open to taking the minimalistic RAUC_CONF variable approach for both master and scarthgap.
> 
> How about we use that minimalistic approach in all of kirkstone, scarthgap and master and additionally change master's default..?

Yes, sounds like a plan.

a3f:
> My comment was about not installing system.conf at all. I guess we won't make this the default.

Ok, I will just drop that.

> Yes, sounds like a plan.

Well, the PR for the first of the bunch is now ready to merge :-)

jluebbe (approved):

PR #351, opened 2024-11-15
bundle.bbclass: more helpful error message for 'file' image type

ejoerns:
We know what's missing, thus let the user know what to do, too.


classes-recipe/bundle.bbclass
```diff
@@ -262,7 +262,7 @@ def write_manifest(d):
         elif imgtype == 'file':
             imgsource = slotflags.get('file')
             if not imgsource:
-                bb.fatal('Unknown file for slot: %s' % slot)
+                bb.fatal('Slot type "file" requires [file] varflag to be set for %s' % slot)
```
jluebbe, line 265: Perhaps
```suggestion
                bb.fatal('Image type "file" requires [file] varflag to be set for slot %s' % slot)
```
?

PR #353, opened 2024-11-15
bundle.bbclass: extend do_bundle[file-checksums]

grembeter:
Rerun `do_bundle` task when `RAUC_CERT_FILE` or `RAUC_KEY_FILE` files have been changed on the disk.

### DIFF
--- a/classes-recipe/bundle.bbclass
+++ b/classes-recipe/bundle.bbclass
@@ -448,6 +448,7 @@ do_bundle() {
 }
 do_bundle[dirs] = "${B}"
 do_bundle[cleandirs] = "${B}"
+do_bundle[file-checksums] += "${RAUC_CERT_FILE}:False ${RAUC_KEY_FILE}:False"
 
 addtask bundle after do_configure
 


ejoerns (changes requested):
This looks really useful. Actually, both files are not part of `SRC_URI` and thus their checksum would not be recorded by default.

However, since (despite the name does not suggest it) both could also be PKCS#11 URIs, I guess using `False` for the boolean part is required here.

grembeter:
Unfortunately, I cannot test it with PKCS#11 URIs.

ejoerns:
> Unfortunately, I cannot test it with PKCS#11 URIs.

@grembeter That shouldn't matter. The point is that the URIs are not valid paths in the file system and thus cannot be tracked. And using `False` allows to have non-existing paths.

Thanks for updating the PR!

ejoerns (approved):

ejoerns:
cherry-picked to [scarthgap](https://github.com/rauc/meta-rauc/commit/d859468ae885fb847733852a9e9b5fdbe946e2fa)

PR #356, opened 2024-11-29
styhead: .github/workflows/build: remove scheduled build

hnez:
Scheduled builds only run for the main branch of a project, so configuring them on other release branches does not make a lot of sense.


ejoerns:
For some more background, I just link #357 here.

ejoerns (approved):

PR #358, opened 2024-11-29
README: update GitHub CI badges

hnez:
The previous badge pointed to the `meta-rauc.yml` workflow file, which is no longer used in any branch.
Instead the workflow now resides in `build.yml` in all branches.

Daily scheduled builds are triggered from `schedule-builds.yml` via `workflow_dispatch` for all branches.
Display the status of the most recent `workflow_dispatch` build for all currently supported branches.

We should likely include the release name in the respective `build.yml` for each branch so it is shown in the badge.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,4 +1,16 @@
-|MIT| |gh_action| |Matrix|
+.. list-table::
+   :header-rows: 1
+
+   * - master
+     - styhead
+     - scarthgap
+     - kirkstone
+   * - |gh_master|
+     - |gh_styhead|
+     - |gh_scarthgap|
+     - |gh_kirkstone|
+
+|MIT| |Matrix|
 
 The meta-rauc layer provides support for integrating the RAUC update tool
 into your device.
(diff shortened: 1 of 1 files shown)


ejoerns:
> We should likely include the release name in the respective `build.yml` for each branch so it is shown in the badge.

Indeed, having the same badge four times looks a little odd. As branch names or customizing the badge does not seem to be supported, I'd also tend to include the branch/release name in the `build.yml`'s `name`.

hnez:
I've just opened another pull request that does basically the same in meta-ptx https://github.com/pengutronix/meta-ptx/pull/166 but with the release names set in the `build.yml`'s `name` field.
We can experiment there first to find out if we like the look.

hnez:
Setting the `name` field did not work out quite like I wanted it to.

I've instead experimented with using a table to display the different build status: [README](https://github.com/hnez/meta-rauc/tree/f947f3e80a9127b8dfb1e616489994905574b552).

What do you think?

ejoerns (approved):
To me, it looks like a good compromise if the alternatives just don't work.

jluebbe (approved):

PR #362, opened 2024-12-16
Add support for full custom update handler

dennisvec:
Rauc allows a script to be defined that will perform a full custom update. This script is referenced in the manifest in a section called [handler]. Allow users to specify a script to be provided for this purpose.


jluebbe:
See https://github.com/rauc/meta-rauc/pull/363 for the current discussion.

PR #363, opened 2024-12-16
Add support for full custom update handler

dennisvec:
Rauc allows a script to be defined that will perform a full custom update. This script is referenced in the manifest in a section called [handler]. Allow users to specify a script to be provided for this purpose.

### DIFF
--- a/classes/bundle.bbclass
+++ b/classes/bundle.bbclass
@@ -12,6 +12,8 @@
 #   RAUC_BUNDLE_HOOKS[file] ?= "hook.sh"
 #   RAUC_BUNDLE_HOOKS[hooks] ?= "install-check"
 #
+#   RAUC_FULL_CUSTOM_HANDLER[file] ?= "custom-handler.sh"
+#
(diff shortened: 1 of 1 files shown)


jluebbe:
I would have expected that a full custom update bundle would be built outside of any build system, as it should be a one-time recovery operation. What is your use-case to generate them via Yocto?

dennisvec:
In our case the slots on the target system are changing quite frequently (we are adding new slots). Changing slots is somewhat difficult to do since the system.conf needs to be upgraded prior to supporting the slot so this either requires a 2 step upgrade, or some out-of-band method to upgrade system.conf and restarting rauc prior to starting the upgrade. To avoid these issues, we just implemented a custom upgrade handler that works without a system.conf upgrade. Our use case is somewhat corner case; however, rauc does support full-custom-update for special situations (recovery, testing, migration), and if you are already using yocto to build bundles, IMO, it makes more sense to continue to do so for a special use case rather than trying to build a custom bundle manually. We use rauc on a number of projects and frequently encounter the situation where one or more slots need to be added at some point during the lifecycle. The full custom update handler provides a reasonably painless way to do this migration.

PR #370, opened 2025-03-31
.github/workflows/build: run on ubuntu 22.04 for python > 3.8.x

ejoerns:
Since walnascar, the minimum required bitbake version will be 3.9.0:

> RuntimeError: Sorry, python 3.9.0 or later is required for this version of bitbake

This will allow continuing to run the GitHub actions on the GH-hosted runners.

Also, explicitly install 'chrpath', which does not seem to be part of the default installation anymore.


ejoerns:
Requires #369 to be merged first.

jluebbe (approved):

PR #372, opened 2025-04-04
bundle.bbclass: make plain bundles compatible with older rauc versions

thwo-bbio:
The bundle section in the manifest was introduced with rauc v1.5. Versions prior to that fail to parse this section. As the section is optional for plain bundles it should be dropped from the manifest to ensure updates can be done even with an old rauc version on the target.

### DIFF
--- a/classes-recipe/bundle.bbclass
+++ b/classes-recipe/bundle.bbclass
@@ -223,14 +223,15 @@ def write_manifest(d):
     manifest.write(d.expand('build=${RAUC_BUNDLE_BUILD}\n'))
     manifest.write('\n')
 
-    if d.getVar('RAUC_BUNDLE_FORMAT'):
-        manifest.write('[bundle]\n')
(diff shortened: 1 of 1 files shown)


thwo-bbio:
Could you please also backport this to langdale, mickledore, nanbield, scarthgap and styhead, if this gets accepted?

ejoerns (approved):

jluebbe (approved):

ejoerns:
cherry-picked to [scarthgap](https://github.com/rauc/meta-rauc/commit/a0f4a8b9986954239850b9d4256c003c91e6b931)

ejoerns:
backported to [kirkstone](https://github.com/rauc/meta-rauc/commit/0b0bfcee1c6c17dff8bad6b4bc69c6f7afd5b827) in a slightly modified version to match the code.

PR #375, opened 2025-05-05
[RFC] bundle.bbclass: unpack flag

Louson:
Add the 'unpack' flag to slots so images have to be unpacked before copied to the bundle directory. Unpacking supports bzip2, gzip, xz and zstd.

For example, using the ext4.zst rootfs:
    RAUC_SLOT_rootfs ?= "core-image-minimal"
    RAUC_SLOT_rootfs[fstype] = "ext4"
    RAUC_SLOT_rootfs[file] = "core-image-minimal-qemuarm64.rootfs.ext4.zst"
    RAUC_SLOT_rootfs[unpack] = "zst"

This patch has been tested in openembedded-core with the above example on core-bundle-minimal and having IMAGE_FSTYPES = "ext4.zst" in the local.conf.

RFC: This currently lacks support for image type 'boot' when having the offset flag. My suggestion would be to use subprocess pipes to send the output of decompression to dd.


jluebbe:
The commit message should describe in which cases this approach (create compressed images and then decompress for RAUC bundles)  is useful. In our projects, we use IMAGE_FSTYPES = "ext4" (or other raw images) to avoid the compression CPU overhead.

Louson:
I have changed the commit message to describe the situation. I agree with the CPU overhead but the goal here is to save disk space and bandwith. Please tell me if this is better

jluebbe:
I'd be fine with not supporting the combination of offset with unpack, as an offset is mainly useful for bootloader images, which are small anyway and don't really need to be compressed.

The new flag should be documented in the initial comment block, mentioning the use-case. Also mention the limitation of not supporting the combination with offset.

ejoerns:
@Louson Agreed, having a way to explicitly set the decompression might be useful, anyway. Extending it with auto-detection for the common extensions can then be an extension.

PR #378, opened 2025-06-25
layer.conf: remove 'styhead' from LAYERSERIES_COMPAT

hnez:
The styhead release has gone out of support as of April 2025.
Time to remove support for it.


ejoerns (approved):
Thanks for the clean up!

PR #376, opened 2025-06-21
layer.conf: add whinlatter to LAYERSERIES_COMPAT for compatibility

Gerry546:
New yocto release is created in poky. This add rauc layer support for that release.

PR #379, opened 2025-06-25
Fix UNPACKDIR for Yocto release Whinlatter

leon-anavi:
This pull request takes the work from pull request #376 by @Gerry546 and adds a couple of additional commits related to fixing issue following the recent change in openembedded-core: https://git.openembedded.org/openembedded-core/commit/?id=46480a5e66747a673041fe4452a0ab14a1736d5e

### DIFF
--- a/recipes-casync/casync/casync_git.bb
+++ b/recipes-casync/casync/casync_git.bb
@@ -12,8 +12,6 @@ SRC_URI = " \
     git://github.com/systemd/casync.git;protocol=https;branch=main \
     "
 
-S = "${WORKDIR}/git"
-
 inherit meson pkgconfig
 
 EXTRA_OEMESON += "-Dselinux=false -Dman=false -Dudevrulesdir=${nonarch_base_libdir}/udev/rules.d/"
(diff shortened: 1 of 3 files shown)


classes-recipe/bundle.bbclass
```diff
@@ -315,13 +314,13 @@ def write_manifest(d):
             else:
                 shutil.copy(searchpath, bundle_imgpath)
         else:
-            searchpath = d.expand("${UNPACKDIR}/%s") % imgsource
+            searchpath = d.expand("${S}/%s") % imgsource
```
ejoerns, line 317: Is it necessary to change this and the following, too? I'd say `UNPACKDIR` is semantically correct, isn't it?

ejoerns (commented):
@Gerry546 @leon-anavi Thank you for the update and fixes!

The CI is still a bit unhappy. Thus, I felt free to apply the sed commands suggested by oe-core tree-wide and stack my commit on top.

classes-recipe/bundle.bbclass
ejoerns: I'd say they are at least changes unrelated to the actual commit's intention and would drop this for now, so we can merge the rest.

CI looks fine now.

ejoerns (approved):

PR #380, opened 2025-07-14
rauc-hawkbit-updater: update to v1.4

Bastian-Krause:
See https://github.com/rauc/rauc-hawkbit-updater/releases/tag/v1.4


ejoerns:
Note: CI breaks due to meta-oe bug, but succeeded in building `rauc-hawkbit-updater`.

PR #385, opened 2025-08-19
bundle.bbclass: create image type in manifest

zentax-dev:
This adds the type for images in the rauc manifest. With the image type the detection is not based on the images file extension anymore, but it has to be explicitely set now.

Depends on #1756


ejoerns:
Now that https://github.com/rauc/rauc/pull/1756 got merged, work on this could continue.

gumulka:
Does it make sense to have hooks, convert and adaptive in the manifest, when the imagetype is an emptyfs?

gumulka:
There is https://github.com/rauc/meta-rauc/blob/3d4a466cd5df98112b0dd70211be7c6a49682e5c/classes-recipe/bundle.bbclass#L333

I feel like this PR should address that line additionally, since it brings in image type support.

classes-recipe/bundle.bbclass
```diff
@@ -300,6 +312,13 @@ def write_manifest(d):
         if 'convert' in slotflags:
             manifest.write("convert=%s\n" % slotflags.get('convert'))
 
+        manifest_imagetype = slotflags.get('imagetype', None)
+        if manifest_imagetype:
+            manifest.write("type=%s\n" % manifest_imagetype)
+            # emptyfs image type can skip image file handling
+            if manifest_imagetype == "emptyfs":
+                continue
+
```
gumulka: Move to line 309, before the check for adaptive and after the check for hooks.

PR #411, opened 2026-02-24
rauc.inc: add config for pkcs11_engine

gumulka:
Add a PACKAGECONFIG option to en-/disable pkcs11_engine named engine, because it disables engine support for openssl, not pkcs11 support. It is enabled by default to have no disruption to previous versions

### DIFF
--- a/recipes-core/rauc/rauc-target.inc
+++ b/recipes-core/rauc/rauc-target.inc
@@ -63,4 +63,4 @@ FILES:${PN}-service = "\
   ${nonarch_libdir}/systemd/catalog \
   "
 
-PACKAGECONFIG ??= "service network streaming json nocreate gpt"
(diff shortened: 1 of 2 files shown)


ejoerns:
> @gumulka I find the commit message quite hard to read. Maybe split up the sentence or at least set "engine" into quotes.

Thank you for splitting this up.

> With the openssl context left out, I find it hard to tell what engine refers to, thus maybe something like openssl-engine for the config name?

I'd assume you intend to stay with this naming since you didn't change it?

> If I'm not mistaken, this disables engine support for native and nativesdk which does not align well with the commit message.

I can't find this comment addressed. Am I mistaken here?

> Maybe also add a short note about the use case for this option.

You don't find this necessary I guess?

ejoerns:
@gumulka  Also please target the `master` branch for new features.

PR #425, opened 2026-04-08
bundle.bbclass: add support for encrypting bundles

NguyenHao-Qualgo:
Add support for encrypting bundles after creating them. This requires the usage of the "crypt" bundle format and a recipient certificate for encrypting the signed bundle manifest. To encrypt the bundle, set RAUC_ENCRYPTED_CRYPT_BUNDLE to 1 and RAUC_RECIPIENTS_CERT_FILE to the path of the recipient certificate, as well as the keyring:

  RAUC_ENCRYPTED_CRYPT_BUNDLE ?= "1"
  RAUC_RECIPIENTS_CERT_FILE ?= "path/to/recipient-certs.pem"
  RAUC_KEYRING_FILE ?= "path/to/ca.cert.pem"

The encrypted bundle will be deployed with the prefix "encrypted-" in the DEPLOY_DIR. Note, that the bundle with the unencrypted manifest is also still being deployed and can be used for manual encryption for other recipients as well.

### DIFF
--- a/classes-recipe/bundle.bbclass
+++ b/classes-recipe/bundle.bbclass
@@ -76,6 +76,21 @@
 #
 #   RAUC_KEYRING_FILE ?= "ca.cert.pem"
 #
+# To encrypt manifests of 'crypt' format bundles (RAUC_BUNDLE_FORMAT = "crypt")
+# during the BSP build already, set
+#
+#   RAUC_ENCRYPT_CRYPT_BUNDLE = "1"
+#
+# and set a corresponding certificate file of the recipient(s):
+#
+#   RAUC_RECIPIENTS_CERT_FILE ?= "recipient-certs.pem"
+#
+# This requires RAUC_KEYRING_FILE to be set.
+#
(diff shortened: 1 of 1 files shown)


NguyenHao-Qualgo:
Hi, I'd like to backport the crypt bundle format support to the scarthgap branch, since it's LTS release.

ejoerns:
@NguyenHao-Qualgo Since it's an LTS release, I would like to avoid backporting new features. Do you work on a new project that requires this shared key version of RAUC's bundle encryption? If starting new projects, I'd recommend using a recent yocto release. The next LTS will be around in a few weeks.

NguyenHao-Qualgo:
Thanks @ejoerns. Yes, we're working on a new project and have chosen Scarthgap as our starting point.

ejoerns:
There is no benefit und picking an old LTS for new developments. I'd suggest starting with software that isn't 2 years old already (it will be even older when going into production).

NguyenHao-Qualgo:
okay. Thank you. I'll close it

PR #427, opened 2026-04-08
rauc: prepare for hermetic /usr/

### LINKED ISSUE #389: Adapt master/scarthgap for hermetic-usr support
We should port the changes from #350 to `scarthgap` / `master` to allow customizing the config location.

For `master`, we should migrate to `/usr` as default.

ejoerns:
The /etc directory is meant for 'local administrator' usage, which a distro build by OE isn't. For the same reason, several OE upstream packages (including systemd itself) already switch to installing into `/usr` instead of `/etc`.
For details on the concept, see [1].

RAUC supports reading the system.conf from `/usr/` since v1.13 [2].

For the next Yocto LTS release, we adapt to the new installation location, which might require migration for existing integrations. However, for most cases, no manual change should be required.

Add a note to the migration chapter to make users ware of this.

[1] https://0pointer.net/blog/fitting-everything-together.html
[2] https://github.com/rauc/rauc/pull/1557

Partly based on https://github.com/rauc/meta-rauc/commit/4cb7c47686d5dec9bacae6110f1bcaac57568691 ("rauc: prepare support for locating the config in /usr or /run").

Fixes #389

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -37,6 +37,12 @@ To use the optional casync feature with FUSE, you will also need:
 Migration Notes
 ===============
 
+Since **wrynose**, the ``rauc-conf`` recipe installs the ``system.conf`` and
+keyring file into ``/usr/lib/rauc`` to align with the 'hermetic ``/usr/``'
+policy followed by other OE components (like systemd).
+If you rely on the exact path where the ``system.conf`` is installed, you will
+need to adjust your integration.
+
 Since **scarthgap**, the platform configuration (system.conf, keyring, etc.) was
 moved to a separate ``rauc-conf.bb`` recipe to allow building the rauc package
 with ``TUNE_PKGARCH`` and have a clearer separation between the binary and the
@@ -63,7 +69,7 @@ at least the following steps are required:
 
      FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
 
-   The ``rauc-conf`` recipe will install these files into ``/etc/rauc/`` by
+   The ``rauc-conf`` recipe will install these files into ``/usr/lib/rauc/`` by
    default.
 
    For information on how to write a proper RAUC system configuration, please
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -37,6 +37,10 @@ To use the optional casync feature with FUSE, you will also need:
 Migration Notes
 ===============
 
+Since **wrynose**, the ``system.conf`` and keyring file will be installed into
+``/usr/lib/rauc`` by default to align with the 'hermetic ``/usr/``' policy
+followed by other OE components (like systemd), too.
```
jluebbe:
```suggestion
Since **wrynose**, the ``system.conf`` and keyring file will be installed into
``/usr/lib/rauc`` to align with the 'hermetic ``/usr/``' policy
followed by other OE components (like systemd).
```

I'd leave out "by default" as this is not explicitly configurable in our recipe.

recipes-core/rauc/files/system.conf
```diff
@@ -1,6 +1,6 @@
-## This is an example RAUC system configuration. This file will be installed
-## into /etc/rauc/system.conf on your target and describes your system from the
-## perspective of the RAUC update service.
+## This is an example RAUC system configuration. This file will be be looked up
+## in {/etc,/run,/usr/lib}/rauc/system.conf on your target and describes your
```
jluebbe:
```suggestion
## in {/etc,/run,/usr/lib}/rauc/system.conf (in that order) on your target and describes your
```

jluebbe (changes requested):
Looks good otherwise.

README.rst
jluebbe: Perhaps also add something like `If you rely on the exact path where the ``system.conf`` is installed, you will need to adjust your integration.`?

ejoerns: > I'd leave out "by default" as this is not explicitly configurable in our recipe.

But you have the option to use a modified recipe or a custom conf recipe. I changed it to explicitly note that the `rauc-conf` recipe changed.

> Perhaps also add something like If you rely on the exact path where the ``system.conf`` is installed, you will need to adjust your integration.?

Not fully sure if that's needed, but added.

jluebbe (approved):

PR #429, opened 2026-04-20
rauc/rauc-target.inc: add a PACKAGECONFIG for mtd, enabled by default

### LINKED ISSUE #428: Missing runtime dependencies (flash_erase)
Hi,

meta-rauc does not create runtime dependencies properly IMHO. I understand that covering all possible cases for runtime dependencies might not be appropriate but for failures like
```
 40% Checking slot kernel.b (b)
 42% Checking slot kernel.b (b) done.
 42% Copying image to kernel.b
 60% Copying image to kernel.b failed.
 99% Updating slots failed.
100% Installing failed.
LastError: Installation error: Failed updating slot kernel.b: failed to run flash_erase: Failed to execute child process ?flash_erase? (No such file or directory)
```
I know what needs to be done here but it should be simple to add a runtime dependency automatically if one of your system.conf components is using the `type=nor` directive or probably just introduce a new PACKAGECONFIG? Is the user of meta-rauc expected to handle the dependencies manually?

awais11010:
Using `nor` as a artifact's type is a very basic use-case and RAUC requires mtd-utils (flash_erase/flashcp) at runtime to support this. Add a new PACKAGECONFIG to support this while enabling it by default as it is a basic workflow.

Fixes #428

### DIFF
--- a/recipes-core/rauc/rauc.inc
+++ b/recipes-core/rauc/rauc.inc
@@ -22,6 +22,7 @@ PACKAGECONFIG[json]    = "-Djson=enabled,-Djson=disabled,json-glib"
 PACKAGECONFIG[gpt]     = "-Dgpt=enabled,-Dgpt=disabled,util-linux"
 PACKAGECONFIG[composefs] = "-Dcomposefs=enabled,-Dcomposefs=disabled,composefs"
 PACKAGECONFIG[manpages] = "-Dmanpages=true,-Dmanpages=false,python3-sphinx-native python3-sphinx-rtd-theme-native"
+PACKAGECONFIG[mtd]     = ",,,mtd-utils"
 
 FILES:${PN}-dev += "\
   ${datadir}/dbus-1/interfaces/de.pengutronix.rauc.Installer.xml \


jluebbe (changes requested):
Use the term `slot` instead of `artifact`. Artifacts are [something specific](https://rauc.readthedocs.io/en/latest/basic.html#artifact-repositories) in the context of RAUC.

Don't enable it by default, as most new systems use eMMCs and neither NOR nor NAND flashes (at least in our experience across many projects in the recent years). Enabling the `PACKAGECONFIG` is easy in your DISTRO (or MACHINE).

ejoerns:
From the commit message

> Add a new PACKAGECONFIG to support this while which can be enabled via
DISTRO/MACHINE configs.

I don't see that `PACKAGECONFIG` is something that is enabled from the `DISTRO` or `MACHINE` (unless the PACKAGECONFIG itself depends on values in `DISTRO_FEATURES` or `MACHINE_FEATURES`).
How would you use this?

jluebbe:
`PACKAGECONFIG:append:pn-rauc = " mtd"`?

ejoerns:
> `PACKAGECONFIG:append:pn-rauc = " mtd"`?

I know how to work solve (or work around) this technically.

My point was: I find it quite unusual to manipulate individual recipe's `PACKAGECONFIG` entries in a distro or machine config.

In my mental model, this typically works the other way round: `PACKAGECONFIG`s check for `DISTRO_FEATURES` or `MACHINE_FEATURES` set, like:

``` bitbake
PACKAGECONFIG ??= "[other features..] ${@bb.utils.filter('MACHINE_FEATURES', 'mtd', d)}"
```

awais11010:
> Use the term `slot` instead of `artifact`. Artifacts are [something specific](https://rauc.readthedocs.io/en/latest/basic.html#artifact-repositories) in the context of RAUC.
> 
> Don't enable it by default, as most new systems use eMMCs and neither NOR nor NAND flashes (at least in our experience across many projects in the recent years). Enabling the `PACKAGECONFIG` is easy in your DISTRO (or MACHINE).

@ejoerns that's my understanding as well but I was asked to do it this way in the [first comment](https://github.com/rauc/meta-rauc/pull/429#pullrequestreview-4140415008) :)

ejoerns:
> > Use the term `slot` instead of `artifact`. Artifacts are [something specific](https://rauc.readthedocs.io/en/latest/basic.html#artifact-repositories) in the context of RAUC.
> > Don't enable it by default, as most new systems use eMMCs and neither NOR nor NAND flashes (at least in our experience across many projects in the recent years). Enabling the `PACKAGECONFIG` is easy in your DISTRO (or MACHINE).
> 
> @ejoerns that's my understanding as well but I was asked to do it this way in the [first comment](https://github.com/rauc/meta-rauc/pull/429#pullrequestreview-4140415008) :)

I'd agree that it shouldn't be enabled by default.
The feature handling would ensure this (since the feature isn't enabled by defaul).
However, I guess we can take the commit as-is for now and leave it up to the user if and how to enable the config option.
(And in case mtd-related parts of RAUC code become meson-configurable later, we already have an option for it.)

ejoerns (approved):
FTR: Fixed an excess word in the commit message (and added a note on .bbappend).

PR #430, opened 2026-04-23
bundle.bbclass: Fixes for RAUC_BUNDLE_EXTRA_FILES handling

ejoerns:
Changes to make the current handling more robust and fix some minor issues.

### DIFF
--- a/classes-recipe/bundle.bbclass
+++ b/classes-recipe/bundle.bbclass
@@ -475,7 +475,7 @@ python do_configure() {
         # strip leading and trailing slashes to prevent installing into wrong location
         file = file.rstrip('/').lstrip('/')
 
-        if file.find("/") != -1:
+        if "/" in file:
             destdir = file.rsplit("/", 1)[0] + '/'
             bb.utils.mkdirhier("%s/%s" % (bundledir, destdir))
         bb.note("Unpacking %s to %s/" % (file, bundledir))
--- a/classes-recipe/bundle.bbclass
+++ b/classes-recipe/bundle.bbclass
@@ -479,7 +479,7 @@ python do_configure() {
             destdir = file.rsplit("/", 1)[0] + '/'
             bb.utils.mkdirhier("%s/%s" % (bundledir, destdir))
         bb.note("Unpacking %s to %s/" % (file, bundledir))
-        ret = subprocess.call('cp -fpPRH "%s" "%s"' % (searchpath, destdir), shell=True, cwd=bundledir)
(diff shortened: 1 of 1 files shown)


classes-recipe/bundle.bbclass
```diff
@@ -479,7 +479,7 @@ python do_configure() {
             destdir = file.rsplit("/", 1)[0] + '/'
             bb.utils.mkdirhier("%s/%s" % (bundledir, destdir))
         bb.note("Unpacking %s to %s/" % (file, bundledir))
-        ret = subprocess.call('cp -fpPRH "%s" "%s"' % (searchpath, destdir), shell=True, cwd=bundledir)
+        subprocess.check_call('cp -fpPRH "%s" "%s"' % (searchpath, destdir), shell=True, cwd=bundledir)
```
hnez, line 482: The string escaping (or lack thereoff) gives me the heebie-jeebies. I guess there is a reason why `call` / `check_call` are used here and why `shell=True` is set, but just in case there is not, how about `subprocess.run`?

```suggestion
        subprocess.run(['cp', '-fpPRH', searchpath, destdir], cwd=bundledir, check=True)
```

classes-recipe/bundle.bbclass
```diff
@@ -475,7 +475,7 @@ python do_configure() {
         # strip leading and trailing slashes to prevent installing into wrong location
         file = file.rstrip('/').lstrip('/')
 
-        if file.find("/") != -1:
+        if "/" in file:
             destdir = file.rsplit("/", 1)[0] + '/'
             bb.utils.mkdirhier("%s/%s" % (bundledir, destdir))
```
hnez, line 480: These do sound like a job for `os.path.dirname` and `os.path.join` instead, but since this commit is already an incremental improvement I do not insist on rewriting the rest as well (at the risk of breaking something).

classes-recipe/bundle.bbclass
ejoerns: I actually intended to perform major changes on the code, but got tangled up with the actual rework and ended up with first of all fixing the very low hanging fruits.
This is why I first of all wanted to change only what's necessary. But I'd keep this in mind for a future rework.

classes-recipe/bundle.bbclass
ejoerns: I also had a version with Pathlib, but since this is both unrelated to the changes done so far, I'd prefer addressing this in a follow-up rework/cleanup.

hnez (approved):
These changes are valid incremental improvements, so I approve of merging them as is.

PR #431, opened 2026-05-06
bundle.bbclass: Fix and Improve parts of the inline bundle usage documentation

### LINKED ISSUE #368: RAUC_SLOT_rootfs[rename]
Hi,

While working on the RAUC integration for  STM32MP157F-DK2 and other STM32MP embedded Linux devices, I had to set `IMAGE_MACHINE_SUFFIX="-${DISTRO}-stm32mp15-disco"` in the RAUC bundle.

Now I am trying to use `RAUC_SLOT_rootfs[rename]` instead but I am running into a behavior that looks like a bug in class `classes-recipe/bundle.bbclass` from layer meta-rauc. In the recipe I have set:

```
RAUC_SLOT_rootfs[rename] = "st-image-core-${DISTRO}-${MACHINE}.rootfs.ext4"
```
The result is the following note and error:

```
NOTE: adding image to bundle dir: 'st-image-core-openstlinux-weston-stm32mp15-disco.rootfs.ext4'
ERROR: Failed to find source st-image-core-stm32mp15-disco.rootfs.ext4
```

In `bundle.bbclass` variable `imgname` is used to print the NOTE and the value matches `RAUC_SLOT_rootfs[rename]`. But after that for the ERROR variable `imgsource` is used. Variable `imgsource` has a different value from `RAUC_SLOT_rootfs[rename]` and as a result it points to a file that doesn't exist in the particular case.

I am testing on Scarthgap but this particular code in the `bundle.bbclass` looks the same in master.

Am I missing something or doing anything wrong with `RAUC_SLOT_rootfs[rename]` ?

ejoerns:
* Rewrites the inline comment-based introduction for bundle.bbclass
* Clarifies 'rename' and 'name' varflag handling

Fixes: #368

### DIFF
--- a/classes-recipe/bundle.bbclass
+++ b/classes-recipe/bundle.bbclass
@@ -1,34 +1,37 @@
-# Class for creating rauc bundles
+# Class for creating RAUC bundles
 #
-# Description:
-# 
-# You have to set the slot images in your recipe file following this example:
+# See https://rauc.readthedocs.io/en/latest/integration.html for general
+# information on how to integrate RAUC into your platform.
+#
+# An example bundle recipe can look as follows:
+#
+#   inherit bundle
+#   [...]
 #
 #   RAUC_BUNDLE_COMPATIBLE ?= "My Super Product"
-#   RAUC_BUNDLE_VERSION ?= "v2015-06-07-1"
 #
-#   SRC_URI += "file://hook.sh"
+#   RAUC_BUNDLE_FORMAT = "verity"
 #
-#   RAUC_BUNDLE_HOOKS[file] ?= "hook.sh"
-#   RAUC_BUNDLE_HOOKS[hooks] ?= "install-check"
-#
-#   RAUC_BUNDLE_SLOTS ?= "rootfs kernel dtb bootloader"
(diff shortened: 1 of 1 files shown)


classes-recipe/bundle.bbclass
```diff
@@ -1,34 +1,37 @@
...
+# An example bundle recipe can look as follows:
+#
+#   inherit bundle
+#   [...]
 #
-#   SRC_URI += "file://hook.sh"
+#   RAUC_BUNDLE_COMPATIBLE ?= "My Super Product"
 #
-#   RAUC_BUNDLE_HOOKS[file] ?= "hook.sh"
-#   RAUC_BUNDLE_HOOKS[hooks] ?= "install-check"
```
hnez, line 13: With these lines removed we no longer have documentation for `RAUC_BUNDLE_HOOKS` in the top comment.

Maybe we should add a short section somewhere else in the class documentation.

ejoerns: Yes, that's a good hint. Will add a commit with notes about how to set hooks (which is probably a quite common use case) to this PR.

jluebbe (approved):

PR #432, opened 2026-05-07
layer.conf: Correct name for meta-filesystems

trini:
The collections name for meta-filesystems has always been "filesystems-layer" and not "meta-filesystems". Update the line here to match.


ejoerns:
@trini Thanks for addressing this. The actual fix look fine.
There's a double signed-off-by and a merge in your PR branch. Could you fix that?

trini:
> @trini Thanks for addressing this. The actual fix look fine. There's a double signed-off-by and a merge in your PR branch. Could you fix that?

Ah whoops, yes. I tried doing this with web tools instead of my usual cli.

ejoerns (approved):

PR #435, opened 2026-06-02
bundle.bbclass: Improve and document `[depends]` varflag handling

ejoerns:
While this is rarely used, it should at least be documented properly.

Also fix 'unpack' `DEPENDS` handling if 'depends' is set, which would have been skipped before.

### DIFF
--- a/classes-recipe/bundle.bbclass
+++ b/classes-recipe/bundle.bbclass
@@ -203,15 +203,14 @@ python __anonymous() {
         depends = slotflags.get('depends')
         if depends:
             d.appendVarFlag('do_unpack', 'depends', ' ' + depends)
-            continue
-
-        if imgtype == 'image':
-            d.appendVarFlag('do_unpack', 'depends', ' ' + image + ':do_image_complete')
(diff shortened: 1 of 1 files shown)


classes-recipe/bundle.bbclass
```diff
@@ -83,6 +64,12 @@
 # If the offset is negative, bytes will not be added, but removed.
 #   RAUC_SLOT_bootloader[offset] ?= "0"
 #
+# For most cases, the class will automatically add dependencies to recipes
+# providing the slot content. For cases where the default mechanism fails, you
+# can use the 'depends' varflag to add custom dependencies:
```
OnkelUlla: The class adds **either** the default dependencies constructed with the help of the image type **or** the explicitly specified dependencies. Let's add this information to the help text:

```suggestion
# can use the 'depends' varflag to add custom dependencies instead:
```

OnkelUlla (changes requested):
Good catch!

OnkelUlla (approved):
Thanks for adjusting the documentation, looks good!

PR #438, opened 2026-06-08
Replace whinlatter by wrynose badge

ejoerns:


README.rst
```diff
@@ -2,10 +2,10 @@
    :header-rows: 1
 
    * - master
-     - whinlatter
+     - walnascar
```
hnez:
```suggestion
     - wrynose
```

hnez (changes requested):

README.rst
ejoerns: No one will ever notice that I had 'walnascar' in each line before ;) Glad that the era of 'w' OE releases ends now.

hnez (approved):

PR #440, opened 2026-06-23
dt-utils: Allow downloads from Azure - use https

hcl-dr:
The http SRC_URI hits a firewall rule on Azure VM
builds. Switch it to https.


ejoerns:
@hcl-dr I've change the target branch to 'master'. We can cherry-pick this to the `wrynose` stable branch then.

ejoerns (approved):

PR #441, opened 2026-07-10
rauc: allow RDEPENDS on dbus-broker via new VIRTUAL-RUNTIME_dbus variable

a3f:
The dbus-broker project is an alternative broker to the reference dbus implementation that is exclusively written for Linux systems, and makes use of many modern features provided by recent linux kernel releases.

Its use is transparent to RAUC, but the recipe currently hardcodes for the service a runtime dependency on `dbus`, which is the reference implementation.

The Whinlatter release introduced `${VIRTUAL-RUNTIME_dbus}`, which expands to the dbus broker being used.

Switch our recipe to use that. The build-time dependency on `dbus` remains as `dbus-broker` is, as the name implies, only a dbus broker.


a3f:
Would also appreciate a backport to Wrynose once this (or something equivalent) is merged :>

ejoerns (approved):

